PR #217, opened 2024-10-03
feat: Integrate Photo Compression to webp

ice-kreios:
### What does this PR do?
This PR contains service to compress photo to webp. It runs on native pool so we do not need to run in isolate.

Sample Test Result
> flutter: Original size: 1785250 bytes
> flutter: Compressed size: 119004 bytes
> flutter: Compressed to 6.67% of original size

- Add `image_compression_flutter` package to support all platforms
- Create `PhotoCompressService` to use encode photos


ice-kreios:
Closing this one to replace with new one.

PR #245, opened 2024-10-16
feat: request events in stream

ice-orion:
### What does this PR do?
This PR adds global nostr events cache storage, convenient methods for sending / requesting events from nostr

* [nostr-dart lib](https://github.com/ice-blockchain/nostr-dart) v0.0.5 that returns `Stream<EventMessage>` from the `requestMessages` function;
* Adds `nostrCacheProvider` for global cache
* Adds `CacheableEvent` mixin that should be implemented by event to be able to be cached
* Add `send`, `receive` methods to `NostrNotifier` + handling user relays there
* Handle received events as Stream

### DIFF
--- a/lib/app/features/nostr/providers/nostr_notifier.dart
+++ b/lib/app/features/nostr/providers/nostr_notifier.dart
@@ -102,6 +102,12 @@ class NostrNotifier extends _$NostrNotifier {
       ref.read(nostrCacheProvider.notifier).cache(userRelays);
       return userRelays;
     }
+    //TODO:
+    // else
+    // request to identity->get-user for the provided `pubkey` when implemented
+    // and return them here if found:
+    // ref.read(nostrCacheProvider.notifier).cache(userRelays);
+    // return ...
 
     throw UserRelaysNotFoundException();
   }


ice-kreios (dismissed):

lib/app/features/nostr/providers/nostr_notifier.dart
```diff
@@ -13,23 +17,95 @@ class NostrNotifier extends _$NostrNotifier {
...
-    }
+  Future<void> sendOne(
+    EventMessage event, {
+    ActionSource actionSource = const ActionSourceCurrentUser(),
+  }) async {
+    return send([event], actionSource: actionSource);
+  }
 
-    final relay = await ref.read(relayProvider(userRelays.list.random.url).future);
+  Future<Stream<EventMessage>> request(
```
ice-alcides, line 37: need to return either `Future` or `Stream`, but not `Future<Stream>`

ice-alcides: So it can be like this:

```
Stream<EventMessage> request(
    RequestMessage requestMessage, {
    ActionSource actionSource = const ActionSourceCurrentUser(),
  }) async* {
      final relay = await _getRelay(actionSource);
      yield* requestEvents(requestMessage, relay);
```

lib/app/features/nostr/providers/nostr_cache.dart
```diff
@@ -0,0 +1,31 @@
...
+
+  void cache(CacheableEvent event) {
+    state = {
+      ...state,
+      event.cacheType: {...(state[event.cacheType] ?? {}), event.cacheKey: event},
+    };
+  }
+}
+
+AlwaysAliveProviderListenable<T?> nostrCache<T>(String key) {
```
ice-alcides, line 29: This approach can lead to unexpected behavior and potential memory leaks (https://riverpod.dev/docs/essentials/do_dont)

ice-alcides (dismissed):

lib/app/features/nostr/providers/nostr_cache.dart
ice-alcides: looks good for me

ice-alcides (approved):

ice-kreios (approved):

PR #249, opened 2024-10-18
feat(auth): add refreshing tokens flow

ice-tychon:
### What does this PR do?
This PR adds a feature to refresh tokens, so a user is not logged out after their token expires.

- The main change is `AuthInterceptor `, it catches 401 errors from BE, blocks other requests (because it's a `QueuedInterceptor`) and tries to refresh token. After that it sets a new token header.

- Added `DelegatedLoginService` with its dependencies to ion_identity_client. This thing makes the request to get a new token.
- Updated ion_identity_client/example a little bit.
- Added `AppLifecycleProvider`, it holds the `AppLifecycleState`

### DIFF
--- a/lib/app/features/core/providers/app_lifecycle_provider.dart
+++ b/lib/app/features/core/providers/app_lifecycle_provider.dart
@@ -1,0 +1,18 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:flutter/material.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'app_lifecycle_provider.g.dart';
+
+@Riverpod(keepAlive: true)
+class AppLifecycle extends _$AppLifecycle {
+  @override
+  AppLifecycleState build() {
+    return AppLifecycleState.resumed;
+  }
+
+  set newState(AppLifecycleState newState) {
+    state = newState;
+  }
+}
--- a/packages/ion_identity_client/lib/src/auth/services/register_service.dart
+++ b/packages/ion_identity_client/lib/src/auth/services/register_service.dart
@@ -45,10 +45,12 @@ class RegisterService {
           ),
         )
         .flatMap(
-          (r) => tokenStorage.setToken(
-            username: username,
-            newToken: r.authentication.token,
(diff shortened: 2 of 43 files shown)


lib/app/features/core/providers/app_lifecycle_provider.dart
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:flutter/material.dart';
```
ice-alcides, line 3: Providers should ideally not have UI imports. On the other hand, `AppLifecycleState` is just an enum. Let's decide if that's okay for us

lib/app/features/auth/services/refresh_tokens_service.dart
```diff
@@ -0,0 +1,72 @@
...
+      _scheduleTokenRefresh(newToken);
+    } catch (e) {
+      // Retry after a short delay
+      Timer(const Duration(minutes: 1), () => _refreshToken(token));
+    }
+  }
+
+  void dispose() {
+    for (final timer in _refreshTimers.values) {
+      timer.cancel();
```
ice-alcides, line 68: need to cheks if the `timer` is not null

lib/app/features/core/providers/app_lifecycle_provider.dart
ice-tychon: Discussed.

ice-alcides (dismissed):

packages/ion_identity_client/lib/src/auth/services/register_service.dart
```diff
@@ -45,10 +45,12 @@ class RegisterService {
...
         )
         .flatMap(
-          (r) => tokenStorage.setToken(
-            username: username,
-            newToken: r.authentication.token,
-            onError: UnknownRegisterUserFailure.new,
+          (r) => TaskEither.tryCatch(
+            () => tokenStorage.setTokens(
+              username: username,
+              newTokens: r.authentication,
```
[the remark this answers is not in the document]
ice-tychon: Tested, now it works.

ice-alcides (approved):

PR #262, opened 2024-10-23
feat(wallets): add getting wallets from api

ice-tychon:
### What does this PR do?
This PR integrates ION Identity Client Wallets API into the application. Unfortunately, it doesn't have all the information that we'd like to display. Currently it's replaced with "NULL" and other values.

- Made a lot of providers Future/Async
- Updated `nftsData` and `coinsData` providers to use the ION Identity Client
 * there area chains of calls to these providers, so other providers and widgets are updated too
- Updated tests
 * also some tests were removed because they don't make sense now (like create/update/delete wallets, they test fake behaviour) or used mocked data intentionally.


lib/app/features/wallets/pages/edit_wallet_modal/edit_wallet_modal.dart
```diff
@@ -65,10 +70,7 @@ class EditWalletModal extends HookConsumerWidget {
               child: isNameChanged
                   ? Button(
                       onPressed: () {
-                        ref
```
ice-alcides, line 68: Maybe we need to keep methods like `addWallet' and `updateWallet` but with an empty body if you need another implementation of those methods and not remove them from the UI.

ice-alcides (dismissed):

lib/app/features/wallets/pages/edit_wallet_modal/edit_wallet_modal.dart
ice-tychon: Discussed, resolving.

lib/app/features/wallets/pages/wallets_modal/components/wallets_list/wallets_list.dart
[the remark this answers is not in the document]
ice-tychon: Updated. Done.

ice-kreios (approved):

ice-alcides (approved):

PR #337, opened 2024-11-26
feat: support opus codec

ice-kreios:
## Description
- Added Opus audio codec compression functionality
- Implemented WAV audio compression
- Streamlined video/image compression with FFmpeg Kit direct integration
- Removed deprecated compression methods and dependencies
- WAV conversion is required as audio waveforms cannot be generated from Opus format

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/services/audio_wave_playback_service/audio_wave_playback_service.dart
+++ b/lib/app/services/audio_wave_playback_service/audio_wave_playback_service.dart
@@ -7,13 +7,15 @@ import 'package:file_saver/file_saver.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/extensions/extensions.dart';
 import 'package:ion/app/services/logger/logger.dart';
+import 'package:ion/app/services/media_service/media_compress_service.dart';
 import 'package:path_provider/path_provider.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'audio_wave_playback_service.g.dart';
 
 class AudioWavePlaybackService {
-  const AudioWavePlaybackService();
+  const AudioWavePlaybackService(this._mediaCompressionService);
+  final MediaCompressionService _mediaCompressionService;
 
   Future<void> initializePlayer(
     String id,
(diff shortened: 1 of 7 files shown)


lib/app/services/media_service/media_compress_service.dart
```diff
@@ -148,93 +93,126 @@ class MediaCompressionService {
...
+  }
 
-      if (decodedImage == null) {
-        throw Exception('Failed to decode image.');
+  ///
+  /// Compress an audio to wav format.
+  /// If success, returns the path to the compressed audio file.
+  /// If fails, throws an exception.
+  ///
+  Future<String> compressAudioToWav(String inputPath) async {
```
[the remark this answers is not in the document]
ice-kreios: because we need to generate a wave form of audio files. however it did not work with opus codec/compression.

lib/app/services/audio_wave_playback_service/audio_wave_playback_service.dart
```diff
@@ -60,9 +62,11 @@ class AudioWavePlaybackService {
...
-    return savedFilePath;
+    final compressedFilePath = await _mediaCompressionService.compressAudioToWav(savedFilePath);
+    return compressedFilePath;
   }
 }
 
 @riverpod
-AudioWavePlaybackService audioWavePlaybackService(Ref ref) => const AudioWavePlaybackService();
+AudioWavePlaybackService audioWavePlaybackService(Ref ref) =>
+    AudioWavePlaybackService(ref.read(mediaCompressServiceProvider));
```
ice-tychon, line 72: Shouldn't it be `ref.watch`?

ice-alcides (approved):

lib/app/services/media_service/media_compress_service.dart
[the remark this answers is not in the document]
ice-kreios: no it is just for local. while sending we should work only with opus

lib/app/services/audio_wave_playback_service/audio_wave_playback_service.dart
ice-kreios: why do you think we should use `watch`?

ice-alcides: It's a recommended way to use `ref.watch `in this type of provider and according to Riverpod's documentation - https://riverpod.dev/docs/essentials/combining_requests#the-refwatch-method

ice-alcides (approved):

PR #340, opened 2024-11-27
fix: stories flow

ice-alcides:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/gallery/providers/camera_provider.dart
+++ b/lib/app/features/gallery/providers/camera_provider.dart
@@ -180,6 +180,18 @@ class CameraControllerNotifier extends _$CameraControllerNotifier {
     );
   }
 
+  Future<XFile?> takePicture() async {
+    if (_cameraController == null) return null;
+
+    try {
+      final picture = await _cameraController!.takePicture();
+      return picture;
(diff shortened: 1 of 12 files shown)


lib/app/features/gallery/providers/camera_provider.dart
```diff
@@ -180,6 +180,18 @@ class CameraControllerNotifier extends _$CameraControllerNotifier {
...
   }
 
+  Future<XFile?> takePicture() async {
+    if (_cameraController == null) return null;
+
+    try {
+      final picture = await _cameraController!.takePicture();
+      return picture;
+    } catch (e) {
+      Logger.log('Error taking picture: $e');
```
ice-brontes, line 190: I see, that logger can take the e too, maybe it's better to use `Logger.log('Error taking picture', error: e);`

ice-alcides: changed

ice-brontes (approved):

ice-kreios (approved):

PR #421, opened 2024-12-12
fix: prevent long text from overflowing by enabling line wrapping

ice-alcides:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/stories/views/components/story_viewer/components/core/story_input_field.dart
+++ b/lib/app/features/feed/stories/views/components/story_viewer/components/core/story_input_field.dart
@@ -4,10 +4,12 @@ import 'dart:ui';
 
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/extensions/extensions.dart';
+import 'package:ion/app/features/feed/stories/providers/story_pause_provider.c.dart';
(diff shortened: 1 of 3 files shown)


lib/app/features/feed/stories/views/components/story_viewer/components/core/story_input_field.dart
```diff
@@ -18,10 +20,23 @@ class StoryInputField extends HookWidget {
   final double bottomPadding;
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     final isTextNotEmpty = useState(false);
     final focusNode = useFocusNode();
 
+    useEffect(
```
ice-kreios, line 27: `useNodeFocused` would you mind using this hook?

ice-endymion (dismissed):

lib/app/features/feed/stories/views/components/story_viewer/components/core/story_input_field.dart
ice-alcides: It will not work because useNodeFocused returns ValueListenable. In code, it will look like this:

 ```
 final focusNode = useFocusNode();
    final focused = useNodeFocused(focusNode);

    useEffect(
      () {
        ref.read(storyPauseControllerProvider.notifier).paused = focused.value;
        return null;
      },
      [focused.value],
    );
```
Even I replace `useEffect` with `useOnInit` I get the not expected behaviour

ice-kreios (approved):

PR #480, opened 2024-12-27
chore: removed password flow random apperance for devices which suppo…

ice-hades:
## Description
Removed password flow random appearance for devices which support passkey.
Also added clean for generate code script because now it fails without it every time

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [x] Chore

### DIFF
--- a/lib/app/features/auth/views/pages/get_started/get_started.dart
+++ b/lib/app/features/auth/views/pages/get_started/get_started.dart
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: ice License 1.0
 
-import 'dart:math';
-
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


ice-tychon (approved):

ice-kreios:
@ice-hades https://github.com/ice-blockchain/ion-app/blob/master/packages/ion_identity_client/lib/src/signer/passkey_signer.dart#L184
 as we discussed before, should we update `!isDeviceSupported` to `isDeviceSupported` ?

ice-kreios (approved):

ice-brontes (approved):

PR #489, opened 2024-12-31
fix: disable two fa options dropdown when there are no options available

ice-ajax:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/auth/views/pages/recover_user_twofa_page/components/twofa_options_selector_button.dart
+++ b/lib/app/features/auth/views/pages/recover_user_twofa_page/components/twofa_options_selector_button.dart
@@ -12,6 +12,7 @@ class TwoFaOptionsSelectorButton extends StatelessWidget {
     required this.isOpened,
     required this.controller,
     required this.optionIndex,
+    this.enabled = true,
     super.key,
   });
 
@@ -19,6 +20,7 @@ class TwoFaOptionsSelectorButton extends StatelessWidget {
   final ValueNotifier<bool> isOpened;
   final MenuController controller;
   final int optionIndex;
+  final bool enabled;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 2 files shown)


lib/app/features/auth/views/pages/recover_user_twofa_page/components/twofa_options_selector_button.dart
```diff
@@ -12,13 +12,15 @@ class TwoFaOptionsSelectorButton extends StatelessWidget {
     required this.isOpened,
     required this.controller,
     required this.optionIndex,
+    this.enabled = true,
```
ice-tychon, line 15: Do we need to visually set it as `disabled` in case it's disabled?

ice-ajax: discussed

lib/app/features/auth/views/pages/recover_user_twofa_page/components/twofa_options_selector_button.dart
```diff
@@ -49,6 +51,7 @@ class TwoFaOptionsSelectorButton extends StatelessWidget {
       ),
       opened: isOpened.value,
       onPressed: () {
+        if (!enabled) return;
```
ice-endymion, line 54: better set onPressed as `null` instead of return;

ice-kreios (approved):

ice-tychon (approved):

lib/app/features/auth/views/pages/recover_user_twofa_page/components/twofa_options_selector_button.dart
ice-ajax: in the current implementation of `Button.dropdown`, the `onTap` is required, so we can't set it to null.

ice-endymion (approved):

PR #492, opened 2024-12-31
fix: replace unique key with global key to avoid losing focus on post details reply bottom textfield

ice-ajax:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/create_post/views/components/reply_input_field/reply_input_field.dart
+++ b/lib/app/features/feed/create_post/views/components/reply_input_field/reply_input_field.dart
@@ -36,7 +36,7 @@ class ReplyInputField extends HookConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final textEditorController = useQuillController();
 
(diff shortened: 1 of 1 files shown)


lib/app/features/feed/create_post/views/components/reply_input_field/reply_input_field.dart
```diff
@@ -36,7 +36,7 @@ class ReplyInputField extends HookConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final textEditorController = useQuillController();
 
-    final inputContainerKey = useRef(UniqueKey());
+    final inputContainerKey = useRef(GlobalKey());
```
ice-tychon, line 39: You can use `useRef` or `useMemoized` to keep the unique key alive after rebuild.

ice-kreios (approved):

PR #502, opened 2025-01-02
fix: set ready only as true of the preview editor to handle URLs correctly

ice-kreios:
## Description
When handling URLs in the preview editor, the current implementation allows for editing which can lead to incorrect URL processing and potential issues with link navigation.

## Additional Notes
N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

N/A

### DIFF
--- a/lib/app/features/feed/views/components/text_editor/utils/quill.dart
+++ b/lib/app/features/feed/views/components/text_editor/utils/quill.dart
@@ -19,6 +19,7 @@ QuillController decodeArticleContent(String encodedContent) {
     return QuillController(
       document: Document.fromDelta(delta),
       selection: const TextSelection.collapsed(offset: 0),
+      readOnly: true,
     );
   } catch (error) {
     throw QuillParseException(error);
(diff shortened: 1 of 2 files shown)


ice-ajax (approved):

lib/app/features/feed/views/components/text_editor/utils/quill.dart
```diff
@@ -19,6 +19,7 @@ QuillController decodeArticleContent(String encodedContent) {
     return QuillController(
       document: Document.fromDelta(delta),
       selection: const TextSelection.collapsed(offset: 0),
+      readOnly: true,
```
ice-endymion, line 22: If parse throw Exception then `QuillController.basic();` will be used, which means edit will be available.

ice-kreios: fixed. i have tested this case so it just renders a blank editor. instead of adding this empty text editor, i have refactored to render a basic widget.

ice-tychon (approved):

ice-endymion (approved):

ice-ajax (approved):

PR #503, opened 2025-01-03
fix: create group UI bugs

ice-damocles:
## Description
Create group flow fixes:
- Edit button to lower case
- Arrow for "Next" button
- Group picture picker without selector 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/chat/views/components/type_selection_list_item.dart
+++ b/lib/app/features/chat/views/components/type_selection_list_item.dart
@@ -41,8 +41,9 @@ class TypeSelectionListItem<T extends SelectableType> extends StatelessWidget {
                     ),
                     color: colors.attentionBlock,
                   ),
-                  margin:
-                      EdgeInsets.only(top: ListItem.defaultConstraints.minHeight - borderRadius),
+                  margin: EdgeInsets.only(
+                      top:
+                          ListItem.defaultConstraints.minHeight - borderRadius),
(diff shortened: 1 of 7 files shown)


ice-ajax (dismissed):

lib/app/features/chat/views/components/type_selection_list_item.dart
```diff
@@ -51,7 +51,10 @@ class TypeSelectionListItem<T extends SelectableType> extends StatelessWidget {
                   ),
                   child: Text(
                     type.getDescription(context),
-                    style: textStyles.body.copyWith(color: colors.quaternaryText),
+                    style: textStyles.body.copyWith(
+                      color: colors.quaternaryText,
+                      fontWeight: FontWeight.w400,
```
ice-kreios, line 56: We should always use lean `textStyles`. If there’s no match, we should ping the designer to use an existing text theme.

ice-damocles: My bad, will find proper style now

ice-endymion (approved):

ice-alcides (approved):

ice-kreios (approved):

PR #522, opened 2025-01-08
feat: text toolbar styles combine

ice-hector:
## Description
This PR introduces the ability to combine text style

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Screenshot 2025-01-08 at 18 03 13

### DIFF
--- a/lib/app/features/feed/views/components/toolbar_buttons/toolbar_bold_button.dart
+++ b/lib/app/features/feed/views/components/toolbar_buttons/toolbar_bold_button.dart
@@ -15,7 +15,8 @@ class ToolbarBoldButton extends HookWidget {
   @override
   Widget build(BuildContext context) {
     final fontStyles = useTextEditorFontStyles(textEditorController);
-    final styleManager = QuillStyleManager(textEditorController);
+    final styleManager =
+        useMemoized(() => QuillStyleManager(textEditorController), [textEditorController]);
 
     return ActionsToolbarButton(
       icon: Assets.svg.iconPostBoldtextOff,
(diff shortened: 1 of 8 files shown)


ice-kreios (dismissed):

lib/app/features/feed/views/components/text_editor/utils/toggle_styles.dart
```diff
@@ -0,0 +1,86 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:flutter_quill/flutter_quill.dart';
+
+void toggleHeaderStyle(QuillController controller, Attribute<dynamic> headerAttribute) {
+  final currentStyle = controller.getSelectionStyle();
```
ice-endymion, line 6: maybe be wrap it to class and make static func, or any other way to have some kind of namespace.

lib/app/features/feed/views/components/text_editor/utils/quill_style_manager.dart
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:flutter_quill/flutter_quill.dart';
+
+class QuillStyleManager {
```
ice-endymion, line 5: add private constructor `const QuillStyleManager._();` and make declare static funcs, pass QuillController as param, like this:
dart```
 static void toggleHeaderStyle(
    QuillController controller, {
    required Attribute<dynamic> headerAttribute,
  })
```

ice-endymion: So then `QuillStyleManager` is just namespace to call atomic funcs

ice-kreios: i do not think we should use static functions because at future we ll not able to write unit tests

ice-endymion (dismissed):

ice-endymion (approved):

ice-ajax (approved):

ice-kreios (approved):

PR #549, opened 2025-01-15
feat: add app redirect for each authenticator option on authenticator setup page

ice-ajax:
## Additional Notes
external_app_launcher lib currently doesn't work on iOS because of a bug. Until it is fixed, iOS is using an alternative solution.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/protect_account/authenticator/data/model/authenticator_type.dart
```diff
@@ -3,30 +3,62 @@
...
-      AutethenticatorType.google => 'Google Authenticator',
-      AutethenticatorType.micsrosoft => 'Microsoft Authenticator',
-      AutethenticatorType.lastpass => 'LastPass Authenticator',
-      AutethenticatorType.authy => 'Authy',
-      AutethenticatorType.userLockPush => 'UserLock Push',
+      AuthenticatorType.google => 'Google Authenticator',
+      AuthenticatorType.microsoft => 'Microsoft Authenticator',
+      AuthenticatorType.lastpass => 'LastPass Authenticator',
+      AuthenticatorType.authy => 'Authy',
+      AuthenticatorType.userLockPush => 'UserLock Push',
```
ice-endymion, line 19: They are probably the same for all languages, but it may be better to use `i18n` for consistency.

ice-ajax: fixed

ice-endymion (approved):

ice-alcides (approved):

ice-tychon (approved):

PR #574, opened 2025-01-20
fix: fixed startup crash on android 12

ice-hades:
## Description
Fixed startup issue on android 12 and below.
So there are 2 levels of fixing
 1. added GeneratedPluginRegistrant.registerWith(flutterEngine) call to MainActivity to properly register plugins on older android versions
 2. after that ion_screenshot_detector started to cause a crash on older android versions because there is not available Activity.ScreenCaptureCallback class. So I changed the code to make sure it is never referenced on older android versions

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/packages/ion_screenshot_detector/android/src/main/kotlin/io/ion/app/ion_screenshot_detector/IonScreenshotDetectorPlugin.kt
+++ b/packages/ion_screenshot_detector/android/src/main/kotlin/io/ion/app/ion_screenshot_detector/IonScreenshotDetectorPlugin.kt
@@ -43,7 +43,7 @@ class IonScreenshotDetectorPlugin : FlutterPlugin, EventChannel.StreamHandler, A
   }
 
   override fun onDetachedFromActivity() {
-    registerScreenshotDetector()
+    unregisterScreenshotDetector()
     activity = null
   }
 


ice-kreios (dismissed):

ice-ajax (dismissed):

packages/ion_screenshot_detector/android/src/main/kotlin/io/ion/app/ion_screenshot_detector/IonScreenshotDetectorPlugin.kt
```diff
@@ -36,26 +34,26 @@ class IonScreenshotDetectorPlugin: FlutterPlugin, EventChannel.StreamHandler, Ac
...
 
   override fun onAttachedToActivity(binding: ActivityPluginBinding) {
     activity = binding.activity
-    registerScreenCaptureCallback()
+    registerScreenshotDetector()
   }
 
   override fun onDetachedFromActivity() {
-    unregisterScreenCaptureCallback()
+    registerScreenshotDetector()
```
ice-alcides, line 46: are you sure we need to enable a screenshot detector during the `onDetachedFromActivity` callback?

ice-alcides (approved):

ice-ajax (approved):

ice-kreios (approved):

PR #595, opened 2025-01-22
fix: wallets parsing

ice-brontes:
## Description
Made a few fields nullable, so they can be parsed properly. 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/wallets/providers/current_user_wallet_views_provider.c.dart
+++ b/lib/app/features/wallets/providers/current_user_wallet_views_provider.c.dart
@@ -42,8 +42,7 @@ Future<List<WalletViewData>> currentUserWalletViews(Ref ref) async {
             ?.asset;
 
         if (asset != null) {
-          coinAmount = asset.balance;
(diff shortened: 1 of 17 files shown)


lib/app/features/wallets/providers/current_user_wallet_views_provider.c.dart
```diff
@@ -42,8 +42,7 @@ Future<List<WalletViewData>> currentUserWalletViews(Ref ref) async {
             ?.asset;
 
         if (asset != null) {
-          coinAmount = asset.balance;
-          // TODO: (1) Check the result of formula with other services
+          coinAmount = double.tryParse(asset.balance) ?? 0;
```
ice-alcides, line 45: So, in case of failed parsing, do we show the user that he has zero on the balance?

ice-alcides (approved):

ice-tychon (approved):

PR #603, opened 2025-01-24
feat: add created time for the different post entities

ice-alcides:
## Description

- add localized timestamp widget and refactor date formatting

<p><strong>Acceptance Criteria:</strong></p>
<ol>
<li>
<p>Should be allocated by design near the user's nickname.</p>
</li>
<li>
<p>Should be present on posts, articles, videos, quotes, and replies (everything except reposts)</p>
<p>2.1 Reposts should show only the timestamp of the original post.</p>
</li>
<li>
<p>On the Feed, only the timestamp should be shown. However, when opening post details, both the date and time should be displayed (for posts, articles, videos, quotes, and replies).</p>
<p>3.1 In post details the date and time should be displayed as short month-day-year, hh:mm AM/PM (e.g., Jan 1, 2024, 08:45 PM).</p>
<p>3.2 The creation time should be shown in the device's timezone.</p>
</li>
<li>The timestamps should be translatable if user chosen app language is not US.</li>
<li>
<p>Values for timestamps:</p>
</li>
</ol>

Timestamp | Range | Format | Example | Comments
-- | -- | -- | -- | --
Minutes | 1m - 59m | m | 1m, 5m, 45m |  
Hours | 1h - 23h | h | 3h, 12h, 22h | 22:10 is 22h, and 22:45 is also 22h.
Days | 24h - 7d | d | 1d, 3d, 6d | 1d 12h is still 1d, 3d 22h is still 3d.
Weeks | 8d+ | short month-day | Jan 10 |  
Months | 1m-12m | short month-day | Jan 10 |  
Years | > 12m | short month-day-year | Dec 20, 2024 |  

image

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/feed/views/components/timestamp_widget.dart
```diff
@@ -0,0 +1,36 @@
...
+  final DateTime createdAt;
+  final bool showDetailed;
+  final TextStyle? style;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final locale = ref.watch(appLocaleProvider);
+    final formattedTime = showDetailed
+        ? formatDetailedPostTime(createdAt, locale: locale)
+        : formatFeedTimestamp(createdAt, locale: locale);
```
ice-ajax, line 26: could we change a naming here so it's more universal if we want to use it somewhere else than feed?

```suggestion
        : formatShortTimestamp(createdAt, locale: locale);
```

lib/app/features/feed/views/components/user_info/user_info.dart
```diff
@@ -13,12 +14,14 @@ class UserInfo extends HookConsumerWidget {
...
     required this.pubkey,
     this.trailing,
+    this.timestamp,
     this.textStyle,
     super.key,
   });
 
   final String pubkey;
   final Widget? trailing;
+  final Widget? timestamp;
```
ice-ajax, line 24: If I see correctly, you pass the same `TimestampWidget` to `UserInfo` every time. Wouldn't it be better for `UserInfo` widget to just accept a `DateTime? timestamp` and build the timestamp view itself in that case? Seems easier because any time you need to change how the timestamp looks, you would just change it here, in a single place.

lib/app/utils/date.dart
```diff
@@ -141,3 +97,50 @@ DateTime randomDateBefore(Duration maxDuration) {
...
+  final localDateTime = dateTime.toLocal();
+  final now = DateTime.now();
+  final difference = now.difference(localDateTime);
+  final diffInMinutes = difference.inMinutes;
+  final diffInHours = difference.inHours;
+  final diffInDays = difference.inDays;
+
+  if (diffInMinutes < 60) {
+    return '${diffInMinutes}m';
+  }
```
ice-ajax, line 120: What if the post was created 10 seconds ago? It will show `0m`, right? Maybe we should show "`now`" if it's less than one minute?

lib/app/features/feed/views/components/user_info/user_info.dart
ice-alcides: yeah, good point, let me try

lib/app/utils/date.dart
ice-alcides: I'll clarify it

lib/app/features/feed/views/components/timestamp_widget.dart
ice-alcides: renamed

lib/app/utils/date.dart
ice-alcides: discussed: if the post is newer than 1 minute (i.e < 60 seconds), we show "1m" as the minimum value

lib/app/features/feed/views/components/user_info/user_info.dart
ice-alcides: changed

ice-tychon (dismissed):

lib/main.dart
```diff
@@ -23,6 +24,7 @@ const _riverpodLoggerEnabled = false;
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
+  registerTimeagoLocalesForEnum();
```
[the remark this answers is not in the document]
ice-alcides: done

lib/app/features/core/model/timeago_locales.dart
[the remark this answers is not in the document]
ice-alcides: moved

ice-ajax (approved):

PR #610, opened 2025-01-27
fix: nullable symbolGroup handling

ice-brontes:
## Description
It turns out that `symbolGroup` can be null, so we need to handle that case.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/packages/ion_identity_client/lib/src/wallets/services/wallet_views/models/short_wallet_view.c.dart
+++ b/packages/ion_identity_client/lib/src/wallets/services/wallet_views/models/short_wallet_view.c.dart
@@ -11,11 +11,12 @@ class ShortWalletView with _$ShortWalletView {
   const factory ShortWalletView({
     required String name,
     required List<WalletViewCoinData> coins,
-    required List<String> symbolGroups,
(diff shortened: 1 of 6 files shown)


packages/ion_identity_client/lib/src/wallets/services/wallet_views/models/short_wallet_view.c.dart
```diff
@@ -11,11 +11,12 @@ class ShortWalletView with _$ShortWalletView {
   const factory ShortWalletView({
     required String name,
     required List<WalletViewCoinData> coins,
-    required List<String> symbolGroups,
     required String createdAt,
     required String updatedAt,
     required String userId,
     required String id,
+    @JsonKey(defaultValue: [])
+    required List<String> symbolGroups,
```
ice-ajax, line 19: FYI:

this is equivalent to:
```dart
@Default([]) required List<String> symbolGroups,
```

ice-ajax (approved):

PR #620, opened 2025-01-29
fix: icons tappable areas

ice-hector:
## Description
This PR introduces new tappable areas for buttons in feed list item footer

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Screenshot 2025-01-29 at 02 55 24

### DIFF
--- a/lib/app/features/components/entities_list/components/post_list_item.dart
+++ b/lib/app/features/components/entities_list/components/post_list_item.dart
@@ -20,6 +20,7 @@ class PostListItem extends StatelessWidget {
 
     return GestureDetector(
       onTap: () => PostDetailsRoute(eventReference: eventReference.encode()).push<void>(context),
+      behavior: HitTestBehavior.opaque,
       child: ScreenSideOffset.small(
         child: Post(
           eventReference: eventReference,


lib/app/components/counter_items_footer/counter_items_footer.dart
```diff
@@ -31,40 +31,67 @@ class CounterItemsFooter extends HookConsumerWidget {
...
+  });
+
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context) {
+    return Container(
+      alignment: Alignment.center,
+      child: child,
+    );
```
ice-ajax, line 95: This is equivalent to:

```suggestion
    return Center(
      child: child,
    );
```

We should avoid using Container if the same effect can be achieved with simpler widgets, as Container is quite complex and could be expensive.

lib/app/features/components/entities_list/components/post_list_item.dart
```diff
@@ -20,7 +20,6 @@ class PostListItem extends StatelessWidget {
 
     return GestureDetector(
       onTap: () => PostDetailsRoute(eventReference: eventReference.encode()).push<void>(context),
-      behavior: HitTestBehavior.opaque,
```
ice-ajax, line 23: why are we removing this? It will cause some of the cell's areas to be not clickable

ice-kreios (dismissed):

ice-endymion (approved):

ice-alcides (approved):

ice-ajax (dismissed):

ice-tychon (dismissed):

lib/app/features/components/entities_list/components/post_list_item.dart
ice-ajax: I remember adding it because there were places in the cell that were not reactive at all, meaning no post details navigation would happen even though you clicked inside a cell.

ice-tychon (approved):

ice-alcides (approved):

ice-kreios (approved):

PR #622, opened 2025-01-30
fix: add an option for text selection in the details post page

ice-alcides:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

image 
image

### DIFF
--- a/lib/app/features/feed/views/components/post/components/post_body/post_body.dart
+++ b/lib/app/features/feed/views/components/post/components/post_body/post_body.dart
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: ice License 1.0
 
 import 'package:flutter/material.dart';
-import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/components/text_span_builder/hooks/use_text_span_builder.dart';
 import 'package:ion/app/components/text_span_builder/text_span_builder.dart';
@@ -22,7 +21,6 @@ class PostBody extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final isSelectionEnabled = useState(false);
     final postMedia = postEntity.data.media.values.toList();
 
     final textSpanBuilder = useTextSpanBuilder(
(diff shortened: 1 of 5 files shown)


ice-brontes (dismissed):

lib/app/features/feed/views/components/post/components/post_body/post_body.dart
```diff
@@ -37,21 +41,20 @@ class PostBody extends HookConsumerWidget {
       crossAxisAlignment: CrossAxisAlignment.stretch,
       children: [
         if (postText.toPlainText().isNotEmpty)
-          Text.rich(
-            postText,
+          GestureDetector(
+            onLongPress: selectable ? () => isSelectionEnabled.value = true : null,
+            child: isSelectionEnabled.value && selectable
+                ? SelectableText.rich(postText)
+                : Text.rich(postText),
```
ice-ajax, line 48: could you explain why we need an additional `isSelectionEnabled` value notifier here? Can't we just use `SelectableText.rich` if `selectable` is `true` and `Text.rich` otherwise?

ice-kreios: I agree. Also, maybe the naming should be more focused on `text selectable` because `selectable` alone sounds more like the entire post can be selected, rather than just the text.

ice-alcides: Renamed to `isTextSelectable` for better clarity. The text selection should be available only on post details screens and should be activated via long press, in addition to the default double-tap behavior

ice-kreios (dismissed):

lib/app/features/feed/views/components/post/components/post_body/post_body.dart
ice-ajax: But with the current logic the text won't be selectable by double-tap until we long press it, because it will be a regular `Text.rich`, no? Also, `SelectableText` doesn't support long press selection by default?

ice-alcides: agree, changed

ice-kreios (approved):

ice-tychon (approved):

PR #652, opened 2025-02-05
fix: receive coins filtering

ice-brontes:
## Description
1. Replace `getCoinsBySymbolGroup` with `getCoinsByFilters` to use the different number of filters in the future.
2. Commented receive NFTs call to avoid 404 response in the logs, since we use mocked data in any way.
3. Added ION testnet to networks.
4. Refactored wallets provider, so now it can handle added wallet without reloading.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,5 @@ fastlane/keys
 
 assets/ion_identity/coins.json
 assets/ion_identity/coins_version.txt
+.aider*
+.env
--- a/.gitignore
+++ b/.gitignore
@@ -63,4 +63,3 @@ fastlane/keys
 assets/ion_identity/coins.json
 assets/ion_identity/coins_version.txt
 .aider*
-.env
--- a/lib/app/features/core/providers/wallets_provider.c.dart
+++ b/lib/app/features/core/providers/wallets_provider.c.dart
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: ice License 1.0
 
-import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/services/ion_identity/ion_identity_client_provider.c.dart';
 import 'package:ion_identity_client/ion_identity.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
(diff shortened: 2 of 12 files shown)


.gitignore
```diff
@@ -62,3 +62,5 @@ fastlane/keys
 
 assets/ion_identity/coins.json
 assets/ion_identity/coins_version.txt
+.aider*
+.env
```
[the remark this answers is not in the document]
ice-brontes: Yep, sorry. I will remove `.env`. `.aider` is a small AI tool, I want to try. Makes it a bit easier to interact with LLM models, but adds a few files that I'd like to ignore. I can remove it, actually, if it's necessary.

lib/app/features/core/providers/wallets_provider.c.dart
```diff
@@ -1,14 +1,31 @@
...
+  Future<void> addWallet(Wallet wallet) async {
+    final currentWallets = state.valueOrNull ?? [];
+
+    // Only add if not already present
+    if (!currentWallets.any((w) => w.id == wallet.id)) {
+      state = AsyncData([...currentWallets, wallet]);
+    }
+  }
+
+  Future<void> refresh() async {
```
ice-kreios, line 26: i think we do need it, if we want to destructure we can call `ref.invalidate`wdyt?

ice-brontes: Yep, you are right. Removed.

ice-kreios (dismissed):

lib/app/features/wallets/data/coins/database/coins_dao.c.dart
```diff
@@ -59,4 +55,15 @@ class CoinsDao extends DatabaseAccessor<CoinsDatabase> with _$CoinsDaoMixin {
...
         .get();
   }
+
+  Future<List<Coin>> getByFilters({String? symbolGroup, String? symbol}) {
+    final query = select(coinsTable);
+    if (symbolGroup != null) {
+      query.where((row) => row.symbolGroup.lower().equals(symbolGroup.toLowerCase()));
+    }
+    if (symbol != null) {
+      query.where((row) => row.symbol.lower().equals(symbol.toLowerCase()));
```
ice-tychon, line 65: If you provide both `symbolGroup` and `symbol`, the `symbol` branch will override the `symbolGroup` where-statement. Is it intended?

ice-kreios (approved):

ice-tychon (approved):

PR #677, opened 2025-02-10
feat: soft delete posts and articles

ice-orion:
## Description
This PR introduces soft-delete of posts and articles.
Soft delete means that we do not really delete entities but rather removing the content and all relative metadata.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

image

### DIFF
--- a/lib/app/features/feed/views/components/deleted_entity/deleted_entity.dart
+++ b/lib/app/features/feed/views/components/deleted_entity/deleted_entity.dart
@@ -20,7 +20,7 @@ class DeletedEntity extends StatelessWidget {
     super.key,
     double? bottomPadding,
     double? topPadding,
-  })  : topPadding = topPadding ?? 10.0,
+  })  : topPadding = topPadding ?? 10.0.s,
         bottomPadding = bottomPadding ?? 10.0.s;
 
   final DeletedEntityType entityType;


lib/app/features/components/entities_list/entities_list.dart
```diff
@@ -51,23 +53,53 @@ class EntitiesList extends HookWidget {
...
-        ref.watch(userMetadataProvider(entity.masterPubkey, network: false)).valueOrNull;
+    // Subscribing to the entity here instead of passing it as param to get the updates
+    // e.g. when the entity is deleted
+    final entity = ref.watch(ionConnectEntityProvider(eventReference: eventReference)).valueOrNull;
+
+    if (entity == null ||
+        isBlockedOrBlocking(ref, entity) ||
+        isDeleted(ref, entity) ||
+        !hasMetadata(ref, entity)) {
+      return const SizedBox.shrink();
```
ice-kreios, line 78: why do we need to filter deleted items here? because if it is deleted we show a `DeletedEntity` widget

lib/app/features/feed/create_post/providers/create_post_notifier.c.dart
```diff
@@ -94,6 +86,66 @@ class CreatePostNotifier extends _$CreatePostNotifier {
...
+        settings: null,
+      );
+
+      await _sendPostEntities([postData]);
+    });
+  }
+
+  Future<List<IonConnectEntity>?> _sendPostEntities(List<EventSerializable> entitiesData) async {
+    //TODO: check the event json according to notion when defined
+    return ref.read(ionConnectNotifierProvider.notifier).sendEntitiesData(entitiesData);
```
ice-kreios, line 146: ref.watch

ice-endymion (approved):

PR #685, opened 2025-02-11
feat: received message status

ice-damocles:
## Description
Received message status (if all participants sent kind 7 received receipt)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


ice-alcides (approved):

lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
```diff
@@ -81,12 +84,30 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
         giftWrapService: giftWrapService,
         privateKey: eventSigner.privateKey,
       );
+
       if (rumor != null) {
-        if (rumor.tags.any((tag) => tag[0] == CommunityIdentifierTag.tagName)) {
+        if (rumor.tags.any((tag) => tag[0] == CommunityIdentifierTag.tagName) ||
+            rumor.kind == PrivateMessageReactionEntity.kind) {
```
ice-kreios, line 90: `rumor.kind == PrivateMessageReactionEntity.kind)`
i think we do need to check it here because, at 91 we check it already

ice-damocles: This is bug which doesn't allow kind 7 to pass this flow

ice-damocles: MessageReaction

PR #695, opened 2025-02-12
feat: read message status

ice-damocles:
## Description
Read message status

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
+++ b/lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
@@ -10,6 +10,7 @@ import 'package:ion/app/features/chat/e2ee/providers/send_e2ee_message_provider.
 import 'package:ion/app/features/chat/model/database/chat_database.c.dart';
 import 'package:ion/app/features/ion_connect/ion_connect.dart';
 import 'package:ion/app/features/ion_connect/model/action_source.dart';
+import 'package:ion/app/features/ion_connect/model/related_event.c.dart';
 import 'package:ion/app/features/ion_connect/providers/ion_connect_event_signer_provider.c.dart';
 import 'package:ion/app/features/ion_connect/providers/ion_connect_notifier.c.dart';
 import 'package:ion/app/services/ion_connect/ion_connect_gift_wrap_service.c.dart';
@@ -33,9 +34,8 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
       throw EventSignerNotFoundException();
     }
 
-    final latestEventMessageDate = await ref
-        .watch(conversationEventMessageDaoProvider)
-        .getLatestEventMessageDate(PrivateDirectMessageEntity.kind);
+    final latestEventMessageDate =
+        await ref.watch(conversationEventMessageDaoProvider).getLatestEventMessageDate(PrivateDirectMessageEntity.kind);
 
     final sinceDate = latestEventMessageDate?.add(const Duration(days: -2));
 
@@ -93,13 +93,12 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
             await ref.watch(conversationEventMessageDaoProvider).add(rumor);
             await sendE2eeMessageService.sendMessageStatus(rumor, MessageDeliveryStatus.received);
           } else if (rumor.kind == PrivateMessageReactionEntity.kind) {
-            final status = rumor.content == 'received'
+            final status = rumor.content == MessageDeliveryStatus.received.name
                 ? MessageDeliveryStatus.received
                 : MessageDeliveryStatus.read;
             final kind14EventId =
-                rumor.tags.firstWhereOrNull((tags) => tags[0] == 'e')?.elementAtOrNull(1);
-            final kind7MasterPubkey =
-                rumor.tags.firstWhereOrNull((tags) => tags[0] == 'b')?.elementAtOrNull(1);
+                rumor.tags.firstWhereOrNull((tags) => tags[0] == RelatedImmutableEvent.tagName)?.elementAtOrNull(1);
+            final kind7MasterPubkey = rumor.tags.firstWhereOrNull((tags) => tags[0] == 'b')?.elementAtOrNull(1);
 
             if (kind7MasterPubkey == null || kind14EventId == null) {
               return;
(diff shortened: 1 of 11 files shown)


lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
```diff
@@ -91,8 +91,11 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
           if (rumor.kind == PrivateDirectMessageEntity.kind) {
             await ref.watch(conversationDaoProvider).add([rumor]);
             await ref.watch(conversationEventMessageDaoProvider).add(rumor);
-            await sendE2eeMessageService.sendReceivedStatus(rumor);
+            await sendE2eeMessageService.sendMessageStatus(rumor, MessageDeliveryStatus.received);
```
ice-kreios, line 94: i think we should not send received status if it's current user's message. so we can decrease hitting to relay and local database. do you agree with me?

lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
```diff
@@ -91,8 +91,11 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
...
           } else if (rumor.kind == PrivateMessageReactionEntity.kind) {
+            final status = rumor.content == 'received'
```
ice-kreios, line 96: instead of declaring magic strings, we may define it in the enum.wdyt?

ice-alcides (approved):

lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
```diff
@@ -91,8 +91,11 @@ class E2eeMessagesSubscriber extends _$E2eeMessagesSubscriber {
...
+                ? MessageDeliveryStatus.received
+                : MessageDeliveryStatus.read;
             final kind14EventId =
```
ice-kreios, line 99: can we use `RelatedImmutableEvent.fromTag` method?

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -187,9 +187,17 @@ class SendE2eeMessageService {
     }
   }
 
-  Future<void> sendReceivedStatus(EventMessage kind14Rumor) async {
+  Future<void> sendMessageStatus(EventMessage kind14Rumor, MessageDeliveryStatus status) async {
+    final allowedStatus = [MessageDeliveryStatus.received, MessageDeliveryStatus.read];
```
ice-kreios, line 191: lets create const outside of method so for each method run it ll not create this variable.

lib/app/features/chat/e2ee/views/pages/one_to_one_messages_page.dart
```diff
@@ -118,7 +119,18 @@ class _MessagesList extends ConsumerWidget {
       return const E2eeConversationEmptyView();
     }
 
-    final messages = ref.watch(conversationMessagesProvider(conversationId!));
+    final messages = ref.watch(
+      conversationMessagesProvider(conversationId!).select((message) {
```
ice-kreios, line 123: we may run it in `conversationMessagesProvider` to decouple business logic from UI code

lib/app/features/chat/model/database/dao/conversation_message_status_dao.c.dart
```diff
@@ -15,33 +15,69 @@ class ConversationMessageStatusDao extends DatabaseAccessor<ChatDatabase>
...
+        final conversationId = conversationMessageTableData.conversationId;
+
+        final conversationMessageTableDataList = await (select(conversationMessageTable)
+              ..where((table) => table.conversationId.equals(conversationId)))
+            .get();
+
+        final allEventMessagesId =
+            conversationMessageTableDataList.map((e) => e.eventMessageId).toList();
+
+        final eventMessagesBeforeEvent = await (select(eventMessageTable)
```
ice-kreios, line 35: just asking to be sure. are we sure about that if user ll see a message, we will mark as read the previous messages? i thought we should mark as read  only if user ll see in the messages page.

lib/app/features/chat/model/database/dao/conversation_message_status_dao.c.dart
```diff
@@ -15,33 +15,69 @@ class ConversationMessageStatusDao extends DatabaseAccessor<ChatDatabase>
...
+              ..where((table) => table.id.isIn(allEventMessagesId))
+              ..where((table) => table.createdAt.isSmallerThanValue(createdAt)))
```
ice-kreios, line 37: We should add a check to prevent running this function on messages that are already in the expected status. For example, if I have 100 messages marked as read and send a new message, the function will process the previous messages again. We should add a filter to update only the necessary messages

lib/app/features/chat/e2ee/providers/e2ee_messages_subscriber.c.dart
ice-damocles: Lets keep if for consistency of the flow, so we send E2E message to all participants, send message status to all members etc

ice-damocles: image

ice-damocles: Plus we have that mentioned in documentation

lib/app/features/chat/model/database/dao/conversation_message_status_dao.c.dart
ice-damocles: In my understanding it works exactly like this, when you open Telegram all previous messages are read

lib/app/features/chat/model/database/dao/conversation_message_status_dao.c.dart
ice-damocles: Good point

ice-kreios (dismissed):

ice-tychon (dismissed):

ice-kreios (approved):

ice-tychon (approved):

PR #696, opened 2025-02-12
feat: send coins logic

ice-brontes:
## Description
Add send coins functionality.

## Additional Notes
1. Removed old PNG network icons.
2. Added possibility to send coins.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactoring
- [ ] Documentation
- [x] Chore


lib/app/features/wallets/domain/coins/coins_comparator.dart
```diff
@@ -0,0 +1,99 @@
...
+      a.coin.abbreviation,
+      b.coin.abbreviation,
+      a.coin.network.serverName,
+      b.coin.network.serverName,
+    );
+  }
+}
+
+class CoinPriority {
+  final _priorityList = const [
```
ice-endymion, line 70: Maybe use some enum values if we have such, or create Enum with coins to prioritize it. I just don't like the way to use plain strings as coin id, in the future we can miss some coin. Play with enum maybe as a single source of coins.

lib/app/features/wallets/model/network.dart
```diff
@@ -3,54 +3,58 @@
...
-  ton,
-  tonTestnet,
-  tron,
-  tronNile,
-  xrpLedger,
-  xrpLedgerTestnet;
+  ion(isTestnet: false),
+  ionTestnet(isTestnet: true),
+  algorand(isTestnet: false),
+  algorandTestnet(isTestnet: true),
```
ice-endymion, line 9: Why need ion and ionTestnet? isTestnet flag not enough?

lib/app/features/wallets/model/network.dart
```diff
@@ -3,54 +3,58 @@
 import 'package:ion/generated/assets.gen.dart';
 
 enum Network {
```
ice-endymion, line 5: Can we use some single enum for currency and List<Network> as a field to get network? (Same point, tying to understand if we can use some single enum for coins)

packages/ion_identity_client/lib/src/networks/services/get_estimate_fees/models/estimate_fee.c.freezed.dart
```diff
@@ -0,0 +1,324 @@
+// coverage:ignore-file
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
ice-endymion, line 2: Check `.gitignore`, we don't need to push the generated code.

ice-tychon: We push generated code for `ion_identity_client` package.

lib/app/utils/num.dart
```diff
@@ -2,8 +2,14 @@
 
 import 'package:intl/intl.dart';
 
-String formatDouble(double value) {
-  final formatter = NumberFormat('#,##0.00', 'en_US');
+String formatDouble(
```
ice-tychon, line 5: Please, add tests for this function to make it easier to understand and make sure it works correctly.

lib/app/features/wallets/views/pages/send_nft_form.dart
```diff
@@ -32,8 +33,8 @@ class SendNftForm extends ConsumerWidget {
 
     final formController = ref.watch(sendAssetFormControllerProvider(type: CryptoAssetType.nft));
     final notifier = ref.read(sendAssetFormControllerProvider(type: CryptoAssetType.nft).notifier);
-    final selectedNft = formController.selectedNft;
-    final selectedContactPubkey = formController.selectedContactPubkey;
+    final selectedNft = formController.assetData!.as<NftAssetData>()!.nft;
```
ice-tychon, line 36: Should we change `assetData` to be non-nullable?

lib/app/features/wallets/model/network.dart
ice-brontes: It would be enough, if we used here classes, not enum. I've added `isTestnet` field just to avoid a huge getter. And there is can be a case, when we need to make something like this: `switch(network) {...}`. With one network for mainnet and testnet the usage of this enum can be more complex... And testnet and mainnet is a different type of networks actually (for example, we have different "scans" for Ethereum mainnet and testnet), so I decided to implement it like that.

lib/app/features/wallets/model/network.dart
ice-brontes: Hm... We have actually a class to describe coins, `CoinData`. Also, we have info about more than 15k coins, so it would be hard to make an enum for it, I suppose.

lib/app/features/wallets/model/network.dart
ice-brontes: Also, pls note, that it's a temporary solution. In the future, we will have a class that will describe network in the same way as `CoinData` with info from server.

lib/app/features/wallets/domain/coins/coins_comparator.dart
ice-brontes: I could get ids of different coins with that `symbol`s, BUT they can have different `symbolGroup`, `network`, etc. Also, this way will be quite ossified, since during the run of the app new coins with the same `symbol` may be added, and we will not be able to properly sort them.

It's hard to apply `enums` for coins, since it's full-fledged data-model...

lib/app/utils/num.dart
ice-brontes: Done.

lib/app/features/wallets/views/pages/send_nft_form.dart
ice-brontes: Made it non-nullable, but have to add `notInitialized` asset data to cover the initial state.

lib/app/features/wallets/model/network.dart
ice-brontes: UPD. After [this](https://github.com/ice-blockchain/ion-app/pull/711) pr Network now is a class with list of ids (lower cased names). I've just added getter here `isTestnet` with another list of testnet networks.

ice-kreios (approved):

ice-alcides (approved):

lib/app/features/wallets/model/network.dart
```diff
@@ -92,4 +136,33 @@ class Network {
     'xrpledger',
     'xrpledgertestnet',
   };
+
+  static List<Network> get all => Network.allowedNetworkIds.map((e) => Network(id: e)).toList();
```
ice-tychon, line 140: We can do it "static final" to cache it after the 1st invocation.

lib/app/features/wallets/model/network.dart
```diff
@@ -1,15 +1,58 @@
 // SPDX-License-Identifier: ice License 1.0
 
 import 'package:ion/app/exceptions/exceptions.dart';
+import 'package:ion/app/utils/string.dart';
 import 'package:ion/generated/assets.gen.dart';
 
 class Network {
-  const Network({required this.name});
+  Network({required this.id});
```
ice-tychon, line 8: Is that `id` or `name`?

lib/app/features/wallets/domain/coins/coins_comparator.dart
ice-tychon: I suppose we shouldn't access the values directly, so we don't need an enum. Also, we can miss a coin in case of enums too.

lib/app/features/wallets/domain/wallet_views/create_update_wallet_view_request_builder.dart
```diff
@@ -57,16 +57,20 @@ class _CreateUpdateRequestBuilder {
...
+    final networkWithWallet = <String, List<Wallet>>{};
+    for (final wallet in userWallets) {
+      final network = wallet.network.toLowerCase();
+      networkWithWallet.putIfAbsent(network, () => []).add(wallet);
+    }
 
     for (final coin in coins) {
-      final walletId = networkWithWallet[coin.network.name.toLowerCase()]?.id;
+      final wallets = networkWithWallet[coin.network.id.toLowerCase()];
+      final walletId = wallets?.firstWhereOrNull((wallet) => wallet.name == walletView?.id)?.id;
```
ice-tychon, line 73: `wallet.name == walletView.id`. Are we sure the names corresponds each other? Maybe we should rename one of them?

lib/app/features/wallets/domain/wallet_views/wallet_views_service.c.dart
```diff
@@ -139,7 +140,7 @@ class WalletViewsService {
     }
 
     return WalletViewData(
-      coinGroups: coinGroups.values.toList(),
+      coinGroups: coinGroups.values.toList()..sort(CoinsComparator().compareGroups),
```
ice-tychon, line 143: Consider using `.sorted` (from collections lib) instead of `sort` to treat lists as immutable collections.

lib/app/features/wallets/model/network.dart
ice-brontes: I named it `id`, cos it used for identifying the network actually in the app and shouldn't be used in the client-server interaction or in the UI layer.

lib/app/features/wallets/domain/wallet_views/create_update_wallet_view_request_builder.dart
ice-brontes: Yes, @ice-cronus found here an issue, that a user can have several wallets in the same network, so we came up with this solution. I'm not sure that I know why is that so.

lib/app/features/wallets/model/network.dart
ice-brontes: I also added a few doc comments under these fields for a better understanding of their responsibilities.

lib/app/features/wallets/model/network.dart
ice-brontes: Done.

ice-endymion (approved):

ice-kreios (approved):

lib/app/features/wallets/domain/wallet_views/wallet_views_service.c.dart
```diff
@@ -139,7 +140,7 @@ class WalletViewsService {
     }
 
     return WalletViewData(
-      coinGroups: coinGroups.values.toList(),
+      coinGroups: coinGroups.values.toList().sorted(CoinsComparator().compareGroups),
```
ice-tychon, line 143: You don't need `.toList()` before `.sorted()`.

lib/app/features/wallets/providers/synced_coins_by_symbol_group_provider.c.dart
```diff
@@ -0,0 +1,28 @@
...
+
+@riverpod
+Future<List<CoinInWalletData>> syncedCoinsBySymbolGroup(Ref ref, String symbolGroup) async {
+  final service = await ref.watch(coinsServiceProvider.future);
+  final coins = await service.getSyncedCoinsBySymbolGroup(symbolGroup);
+  final walletViewCoins =
+      await ref.watch(currentWalletViewDataProvider.future).then((walletView) => walletView.coins);
+
+  final result = <CoinInWalletData>[];
+  for (final coin in coins) {
```
ice-tychon, line 21: Can be improved. Make the `walletViewCoins` a Set, and do something like:
```
final result = coins
  .where((coin) => walletViewCoinsSet.contains(coin.id))
  .map((coin) => CoinInWalletData(coin: coin))
  .sorted(CoinsComparator().compareCoins);
```

lib/app/features/wallets/views/pages/coins_flow/send_coins/components/send_coins_form.dart
```diff
@@ -7,39 +7,68 @@ import 'package:hooks_riverpod/hooks_riverpod.dart';
...
+          final numValue = double.tryParse(amountController.value.text);
+          if (numValue != null && numValue > maxAmount) {
+            amountController.text = formatCrypto(maxAmount);
+          }
+          notifier.setCoinsAmount(amountController.text);
+        }
+
+        amountController.addListener(listener);
+
+        return () => amountController.removeListener(listener);
```
ice-tychon, line 58: You don't need to dispose in this case. Hooks dispose text controllers, so just add the `amountController` to the useEffect's keys to attach the listener to the controller. Also, you could `useOnInit` with the key to make the code shorter.

ice-tychon (approved):

lib/app/features/wallets/providers/synced_coins_by_symbol_group_provider.c.dart
ice-brontes: I'm not sure. We are not filtering here, actually, to use `where`.  If `walletViewCoins` contains coin with coin id, then we use data from `walletViewCoins`, if not - create a new CoinInWalletData with that coin.

lib/app/features/wallets/views/pages/coins_flow/send_coins/components/send_coins_form.dart
ice-brontes: Good point, fiex.

ice-kreios (dismissed):

ice-kreios (approved):

ice-endymion (approved):

PR #712, opened 2025-02-14
Feature/auth adjustments

ice-endymion:
## Description
Auth enhancements

## Additional Notes
- Add `Completer` for initAuth to prevent any race condition during auth process
- Move onClose listener to dedicated mixin
- Add `invalidate` to `ActiveRelaysProvider`
- Refactor `CoinsSyncWrapper` to provider and put it to `initProvider` to prevent `ionIdentityClient` call before auth

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore


ice-kreios (approved):

lib/app/features/auth/providers/auth_provider.c.dart
```diff
@@ -117,3 +120,11 @@ class CurrentIdentityKeyNameStore extends _$CurrentIdentityKeyNameStore {
     state = AsyncData(identityKeyName);
   }
 }
+
+void onLogout(Ref ref, void Function() callback) {
+  ref.listen(authProvider.select((state) => state.valueOrNull?.isAuthenticated), (prev, next) {
+    if (prev != null && prev == true && next == false) {
```
ice-brontes, line 126: I think, we can skip check on null here `if (prev == true && next  == false)`, no?

ice-brontes (dismissed):
LGTM, great changes! Thanks for the doc comments.

lib/app/features/ion_connect/providers/relays_provider.c.dart
```diff
@@ -13,19 +15,21 @@ typedef RelaysState = Map<String, IonConnectRelay>;
 
 @Riverpod(keepAlive: true)
 class Relay extends _$Relay
```
ice-tychon, line 17: Could be funtion, not class provider.

lib/app/features/ion_connect/providers/relay_auth_provider.c.dart
```diff
@@ -0,0 +1,131 @@
...
+// Keeping the state mutable intentionally
+class RelayAuthState {
+  RelayAuthState({required this.completer});
+
+  Completer<void> completer;
+  String? challenge;
+}
+
+@riverpod
+class RelayAuth extends _$RelayAuth {
```
ice-tychon, line 25: The `completer` part is pretty hard. We can stay in the "not completed" state (`AnonymousRelayAuthException`), and the code which awaits for `handleRelayAuthOnAction` might never get a result. Should we resolve the completer with Error? Should we change the implementation to get rid of the completer?

lib/app/features/auth/providers/auth_provider.c.dart
ice-endymion: We need it here

lib/app/features/ion_connect/providers/relay_auth_provider.c.dart
ice-endymion: The idea behind this is that we cannot use relay until it is authorized. During the auth process we can handle exceptions from relay to understand whats goes wrong. So if we complete completer on error it's not change anything until we impl some retry logic for auth.

ice-brontes (dismissed):

ice-tychon (approved):

ice-kreios (approved):

PR #760, opened 2025-02-24
fix: import token page issues

ice-tychon:
## Description
Consider reviewing this PR by commits (each commits fixes an issue)
This PR contains the following changes:
- "token already exists" dialog: changed button text, title and description
- added "paste" button for address text field
- "import" button is now disabled if the form is invalid (selectedNetwork == null || address.isEmpty)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/services/clipboard/clipboard.dart
+++ b/lib/app/services/clipboard/clipboard.dart
@@ -3,3 +3,8 @@
 import 'package:flutter/services.dart';
 
 void copyToClipboard(String text) => Clipboard.setData(ClipboardData(text: text));
+
+Future<String> pasteFromClipboard() async {
+  final data = await Clipboard.getData(Clipboard.kTextPlain);
+  return data?.text ?? '';
+}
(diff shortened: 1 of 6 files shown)


lib/app/services/clipboard/clipboard.dart
```diff
@@ -3,3 +3,8 @@
 import 'package:flutter/services.dart';
 
 void copyToClipboard(String text) => Clipboard.setData(ClipboardData(text: text));
+
+Future<String> pasteFromClipboard() async {
```
ice-kreios, line 7: a small suggestion: we may rename it to `get` instead of `paste` because it doesn’t actually paste anything inside this function. Does that make sense to you?

ice-kreios (approved):

ice-alcides (approved):

lib/app/services/clipboard/clipboard.dart
ice-tychon: Nice suggestions. Done.

ice-kreios (approved):

ice-alcides (approved):

PR #808, opened 2025-03-05
refactor:  insert dummy e2e event before media upload

ice-damocles:
## Description
- Fix bug for profile message wrapped
- Improve media types upload by adding event with dummy metadata (needed for resend media types)

## Type of Change
- [x] Bug fix
- [x] Refactoring

### DIFF
--- a/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
+++ b/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
@@ -106,13 +106,6 @@ class SendE2eeMessageService {
         throw EventSignerNotFoundException();
       }
 
-      final conversationTags = _generateConversationTags(
-        subject: subject,
-        groupImageTag: groupImageTag,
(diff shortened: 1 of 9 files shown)


ice-alcides (approved):

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -127,6 +120,37 @@ class SendE2eeMessageService {
               throw UserPubkeyNotFoundException(masterPubkey);
             }
 
+            // These are used to for uploading state
+            final conversationTagsWithDummyMediaTags = _generateConversationTags(
```
ice-endymion, line 124: Is there are a way to avoid this dummy tags?

ice-damocles: Yes, I am going to use local file URL because I will need it to resend message

ice-endymion (approved):

PR #818, opened 2025-03-07
fix: ui issues

ice-alcides:
## Description

- fix: updates `StoryContent` to conditionally show/hide elements based on both pause and menu open states
- fix: update photo library permission request and localization
- fix: the selection toolbar 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/core/permissions/views/components/permission_dialogs/permission_request_sheet.dart
+++ b/lib/app/features/core/permissions/views/components/permission_dialogs/permission_request_sheet.dart
@@ -36,9 +36,9 @@ class PermissionRequestSheet extends ConsumerWidget {
                 iconAsset: Assets.svg.walletIconWalletGelleryperm,
               ),
             Permission.camera => InfoCard(
-                title: context.i18n.camera_require_access_title(appName),
-                description: context.i18n.camera_require_access_description(appName),
-                iconAsset: Assets.svg.walletIconWalletCamera,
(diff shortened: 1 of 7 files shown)


lib/app/features/core/permissions/views/components/permission_dialogs/permission_request_sheet.dart
```diff
@@ -36,9 +36,9 @@ class PermissionRequestSheet extends ConsumerWidget {
                 iconAsset: Assets.svg.walletIconWalletGelleryperm,
               ),
             Permission.camera => InfoCard(
-                title: context.i18n.camera_require_access_title(appName),
-                description: context.i18n.camera_require_access_description(appName),
-                iconAsset: Assets.svg.walletIconWalletCamera,
+                title: context.i18n.photo_library_require_access_title(appName),
```
ice-kreios, line 39: Let’s assume the user wants to record a story, so the app will open the camera screen. However, if we show copy text for the photo library, I think it might be confusing. Am I missing something?

ice-alcides: Good question. I'll clarify with the PO. Probably, there was a wrong link on the design

ice-kreios (approved):

ice-ajax (approved):

PR #822, opened 2025-03-07
feat: Resend message with media attachment

ice-damocles:
## Description
- Resend failed message with media
- Renew failed media upload

## Type of Change
- [x] Bug fix
- [x] New feature

### DIFF
--- a/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
+++ b/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
@@ -290,15 +290,17 @@ class SendE2eeMessageService {
     final messageType = entity.data.messageType;
     final mediaUrl = entity.data.primaryMedia?.url;
     final mediaUri = Uri.tryParse(mediaUrl ?? '');
-    final isMediaAttachmentUploaded =
-        (messageType == MessageType.video || messageType == MessageType.audio) &&
-            mediaUri != null &&
-            mediaUri.hasScheme;
+    final isMediaAttachmentUploaded = (messageType == MessageType.video ||
+            messageType == MessageType.audio ||
+            messageType == MessageType.document) &&
+        mediaUri != null &&
+        mediaUri.hasScheme;
 
-    final isMediaAttachmentNotUploaded =
-        (messageType == MessageType.video || messageType == MessageType.audio) &&
-            mediaUri != null &&
-            !mediaUri.hasScheme;
+    final isMediaAttachmentNotUploaded = (messageType == MessageType.video ||
+            messageType == MessageType.audio ||
+            messageType == MessageType.document) &&
+        mediaUri != null &&
+        !mediaUri.hasScheme;
 
     final messageStatuses =
         await conversationMessageStatusDao.messageStatuses(failedMessageEvent.id);
--- a/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
+++ b/lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
@@ -146,8 +146,6 @@ class SendE2eeMessageService {
                 throw UserPubkeyNotFoundException(masterPubkey);
               }
 
-              // We need to insert event message to DB before video is uploaded to
-              // retry if upload failed
               if (masterPubkey == currentUserMasterPubkey) {
                 await conversationDao.add([eventMessageWithoutMedia]);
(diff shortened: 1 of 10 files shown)


lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -112,130 +113,139 @@ class SendE2eeMessageService {
...
 
+        // These are used to for uploading state
+        final conversationTagsWithDummyMediaTags = _generateConversationTags(
+          subject: subject,
+          groupImageTag: groupImageTag,
+          conversationId: conversationId,
+          masterPubkeys: participantsMasterPubkeys,
+        )..add([
+            'imeta',
+            'url ${mediaFiles.first.path}',
```
ice-kreios, line 124: why do we get the first of media? i think it ll create an issue if user ll send a multiple media in one message. do you agree with me?

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -112,130 +113,139 @@ class SendE2eeMessageService {
...
-            if (masterPubkey == currentUserMasterPubkey) {
-              await eventMessageDao.add(eventMessageWithMedia);
+            try {
+              final pubkey = participantsKeysMap[masterPubkey];
+
+              if (pubkey == null) {
+                throw UserPubkeyNotFoundException(masterPubkey);
+              }
+
+              // We need to insert event message to DB before video is uploaded to
```
ice-kreios, line 149: `// We need to insert event message to DB before video is uploaded to` i think it works for all attachments now. should we update/remove this comment to avoid confusing?

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -275,6 +285,21 @@ class SendE2eeMessageService {
   }
 
   Future<void> resendFailedMessage(EventMessage failedMessageEvent) async {
+    final entity = PrivateDirectMessageEntity.fromEventMessage(failedMessageEvent);
+
+    final messageType = entity.data.messageType;
+    final mediaUrl = entity.data.primaryMedia?.url;
+    final mediaUri = Uri.tryParse(mediaUrl ?? '');
+    final isMediaAttachmentUploaded =
+        (messageType == MessageType.video || messageType == MessageType.audio) &&
```
ice-kreios, line 294: i think we should handle it with a generic solution. because we need this logic for document message too.

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
```diff
@@ -286,36 +311,142 @@ class SendE2eeMessageService {
...
-          masterPubkey: masterPubkey,
-          eventMessageId: failedMessageEvent.id,
-          status: masterPubkey == currentUserMasterPubkey
-              ? MessageDeliveryStatus.read
-              : MessageDeliveryStatus.sent,
-        );
-      }),
-    );
+    // If this is message without media or message with media but media was successfully uploaded
+    if (messageType == MessageType.profile ||
```
ice-kreios, line 315: i think we can check that if a message contains a media or not so it ll handle all message types. wdyt?

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
ice-damocles: As for now I took first one cause we don't support multiple media, also based on this respective loader will be shown

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
ice-damocles: Refactored with `final isMessageWithMedia = [
      MessageType.video,
      MessageType.audio,
      MessageType.document,
    ].contains(messageType);`

lib/app/features/chat/e2ee/providers/send_e2ee_message_provider.c.dart
ice-damocles: Fixed with `final isMessageWithMedia = [
      MessageType.video,
      MessageType.audio,
      MessageType.document,
    ].contains(messageType);`

ice-kreios (approved):

ice-alcides (approved):

PR #864, opened 2025-03-19
feat: delete entity confirmation

ice-hades:
## Description
Delete entity confirmation pop up

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/2d1e8b8b-0f13-4c85-85ee-0dee71cb8dbb


lib/app/features/feed/data/models/delete/delete_service.dart
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:ion/app/features/ion_connect/model/event_reference.c.dart';
+
+abstract class DeleteService<T> {
+  ProviderBase<T> get provider;
```
ice-alcides, line 7: avoid dynamically creating providers ([link](https://stackoverflow.com/a/74140840/1300882))

lib/app/features/feed/data/models/delete/delete_service.dart
```diff
@@ -0,0 +1,10 @@
...
+
+  Future<void> delete(WidgetRef ref, EventReference eventReference);
```
ice-alcides, line 9: `WidgetRef` is equivalent to relying on `BuildContext`. If you need to get a provider, then use `Ref`. `Ref` is allowed to be used in business logic and you can get any provider via `ref.read`, but ideally, if interfaces and abstract classes know nothing about external dependencies.

lib/app/features/feed/data/models/delete/delete_confirmation_type.dart
```diff
@@ -0,0 +1,47 @@
...
+
+enum DeleteConfirmationType {
+  story(StoryDeleteService()),
+  video(EntityDeleteService()),
+  post(EntityDeleteService()),
+  article(EntityDeleteService());
+
+  const DeleteConfirmationType(this.service);
+
+  final DeleteService<AsyncValue<void>> service;
```
ice-tychon, line 19: Enums are singletons and you gonna make `DeleteService` be alive forever.

lib/app/features/feed/data/models/delete/delete_service.dart
ice-tychon, line 7: It's always `deleteEntityControllerProvider`

lib/app/features/feed/data/models/delete/delete_confirmation_type.dart
ice-tychon: We use Riverpod to provider dependencies.

ice-alcides (approved):

PR #919, opened 2025-03-27
feat: profile ui updates

ice-hades:
## Description
- Updated profile and edit profile UI;

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots (if a
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-27 at 12 13 04
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-27 at 12 13 22
Simulator Screenshot - iPhone 16 Pro Max - 2025-03-27 at 12 13 31
pplicable)

### DIFF
--- a/lib/app/features/auth/views/components/user_data_inputs/location_input.dart
+++ b/lib/app/features/auth/views/components/user_data_inputs/location_input.dart
@@ -3,6 +3,7 @@
 import 'package:flutter/material.dart';
 import 'package:ion/app/extensions/extensions.dart';
 import 'package:ion/app/features/auth/views/components/user_data_inputs/general_user_data_input.dart';
+import 'package:ion/app/features/user/model/user_metadata.c.dart';
 import 'package:ion/app/utils/validators.dart';
 import 'package:ion/generated/assets.gen.dart';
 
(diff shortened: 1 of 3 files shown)


ice-kreios (approved):

lib/app/features/auth/views/components/user_data_inputs/location_input.dart
```diff
@@ -21,10 +22,18 @@ class LocationInput extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    const maxLength = 30;
```
ice-ajax, line 25: could we add that const to UserMetadata, the same way `postCharacterLimit` is added to ModifiablePostEntity? Just to keep it consistent

lib/app/features/auth/views/components/user_data_inputs/website_input.dart
```diff
@@ -22,6 +22,8 @@ class WebsiteInput extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    const maxLength = 100;
```
ice-ajax, line 25: same here

ice-alcides (approved):

lib/app/features/auth/views/components/user_data_inputs/location_input.dart
ice-ajax: that's a business level limit, we don't want to allow location longer than 30 characters, right?

ice-ajax (approved):

ice-kreios (approved):

ice-alcides (approved):

PR #921, opened 2025-03-27
bugfix: Emoji, archive

ice-damocles:
## Description
- Emoji message multiline
- Archive typo
- Archive tile is not selectable

## Type of Change
- [x] Bug fix

### DIFF
--- a/lib/app/features/chat/views/components/message_items/message_types/emoji_message/emoji_message.dart
+++ b/lib/app/features/chat/views/components/message_items/message_types/emoji_message/emoji_message.dart
@@ -39,9 +39,11 @@ class EmojiMessage extends HookConsumerWidget {
           Row(
             mainAxisSize: MainAxisSize.min,
             children: [
-              Text(
-                entity.data.content,
-                style: context.theme.appTextThemes.headline1.copyWith(height: 1),
+              Expanded(
+                child: Text(
+                  entity.data.content,
+                  style: context.theme.appTextThemes.headline1.copyWith(height: 1),
+                ),
               ),
               MessageMetaData(eventMessage: eventMessage),
             ],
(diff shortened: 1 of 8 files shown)


ice-kreios (dismissed):

lib/app/features/ion_connect/providers/ion_connect_notifier.c.dart
```diff
@@ -38,6 +38,7 @@ class IonConnectNotifier extends _$IonConnectNotifier {
     ActionSource actionSource = const ActionSourceCurrentUser(),
     bool cache = true,
     IonConnectRelay? relay,
+    Duration? timeout,
```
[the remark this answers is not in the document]
ice-damocles: Reverted changes to it so not relevant in scope of this PR, please ask to create separate task

lib/app/features/chat/views/components/message_items/message_types/emoji_message/emoji_message.dart
```diff
@@ -39,9 +39,11 @@ class EmojiMessage extends HookConsumerWidget {
           Row(
             mainAxisSize: MainAxisSize.min,
             children: [
-              Text(
-                entity.data.content,
-                style: context.theme.appTextThemes.headline1.copyWith(height: 1),
+              Expanded(
```
ice-kreios, line 42: I think it’d be good to clarify with the designer and PO: should we show a big emoji if the message contains more than one? In Figma, it’s only large when there’s a single emoji, so if the answer is no, we should treat it like a regular text message instead of an emoji message.

ice-damocles: I will ask, but there is specific regex which check if message is only emojis or contains any letters or words

ice-damocles: True, only one, fixed in regexp

ice-kreios (approved):

ice-brontes (approved):

PR #928, opened 2025-03-28
feat: add logic for media files delete on entity delete

ice-ajax:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/providers/delete_entity_provider.c.dart
+++ b/lib/app/features/feed/providers/delete_entity_provider.c.dart
@@ -17,10 +17,12 @@ import 'package:ion/app/features/feed/providers/feed_posts_data_source_provider.
 import 'package:ion/app/features/feed/providers/user_posts_data_source_provider.c.dart';
 import 'package:ion/app/features/feed/providers/user_videos_data_source_provider.c.dart';
 import 'package:ion/app/features/ion_connect/model/deletion_request.c.dart';
+import 'package:ion/app/features/ion_connect/model/entity_data_with_media_content.dart';
 import 'package:ion/app/features/ion_connect/model/event_reference.c.dart';
 import 'package:ion/app/features/ion_connect/model/ion_connect_entity.dart';
 import 'package:ion/app/features/ion_connect/providers/entities_paged_data_provider.c.dart';
 import 'package:ion/app/features/ion_connect/providers/ion_connect_cache.c.dart';
+import 'package:ion/app/features/ion_connect/providers/ion_connect_delete_file_notifier.c.dart';
 import 'package:ion/app/features/ion_connect/providers/ion_connect_entity_provider.c.dart';
 import 'package:ion/app/features/ion_connect/providers/ion_connect_notifier.c.dart';
 import 'package:ion/app/services/logger/logger.dart';
@@ -48,6 +50,7 @@ class DeleteEntityController extends _$DeleteEntityController {
         case GenericRepostEntity() || RepostEntity() || PostEntity():
           {
             await _deleteFromServer(ref, entity);
+            if (entity is PostEntity) _deleteMedia(ref, entity.data);
             _deleteFromDataSources(ref, entity);
             _deleteFromCache(ref, entity);
             _deleteFromCounters(ref, entity);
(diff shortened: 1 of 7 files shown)


ice-endymion (approved):

lib/app/features/ion_connect/providers/ion_connect_delete_file_notifier.c.dart
```diff
@@ -0,0 +1,139 @@
...
+
+      final response = await ref.read(dioProvider).getUri<dynamic>(metadataUri);
+      final deletePath =
+          FileStorageMetadata.fromJson(json.decode(response.data as String) as Map<String, dynamic>)
+              .apiUrl;
+      return metadataUri.replace(path: deletePath).toString();
+    } catch (error) {
+      throw GetFileStorageUrlException(error);
+    }
+  }
```
ice-kreios, line 103: Would it make sense to create shared raw function providers to prevent duplicated logic?

lib/app/features/feed/providers/delete_entity_provider.c.dart
```diff
@@ -149,3 +154,9 @@ void _deleteFromProviders(Ref ref, IonConnectEntity entity) {
     ref.read(repostedEventsNotifierProvider.notifier).removeRepost(eventReference);
   }
 }
+
+void _deleteMedia(Ref ref, EntityDataWithMediaContent entity) {
+  final media = entity.media.values;
```
ice-kreios, line 159: a small suggestion: since it holds a list, maybe we should rename it to `medias` instead of `media`.

ice-ajax: media is both singular and plural in english 😄

lib/app/features/ion_connect/providers/ion_connect_delete_file_notifier.c.dart
ice-ajax: done

lib/app/features/feed/providers/delete_entity_provider.c.dart
ice-kreios: Next time I’ll run my variable names through Grammarly 😄

lib/app/features/ion_connect/providers/ion_connect_delete_file_notifier.c.dart
```diff
@@ -0,0 +1,93 @@
...
+import 'package:ion/app/features/core/providers/dio_provider.c.dart';
+import 'package:ion/app/features/ion_connect/ion_connect.dart';
+import 'package:ion/app/features/ion_connect/utils/file_storage_utils.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'ion_connect_delete_file_notifier.c.freezed.dart';
+part 'ion_connect_delete_file_notifier.c.g.dart';
+
+@riverpod
+class IonConnectDeleteFileNotifier extends _$IonConnectDeleteFileNotifier {
```
ice-alcides, line 17: Can you consider creating a `DeleteFileHandler (Manager/Service)` instead of `Notifier` because this provider doesn't hold any state?

ice-kreios (approved):

ice-endymion (approved):

lib/app/features/ion_connect/providers/ion_connect_delete_file_notifier.c.dart
ice-ajax: I created it for consistency with how the notifier for upload is done. I can refactor both of them in a separate PR if we'd like

ice-ajax: Also I think it is a notifier because it accesses other providers, so it needs a `Ref`

ice-alcides: It is acceptable to pass `Ref ref `as a class parameter or, even better, just `Dio dio`

ice-alcides: So it would look like this:

```
@riverpod
class MyDeletedFileHandler deleteFile (Ref ref) {
  final dio = ref.watch(dioProvider);

  return MyDeleteFileHandler(dio);
  // return MyDeleteFileHandler(ref);
}
```

ice-alcides (approved):

PR #941, opened 2025-03-31
fix: add platform-level error handling

ice-alcides:
## Description

- add error handling for Flutter and platform-level errors in `main.dart`
- ensure context is not null before scrolling in post form modal
- update Talker

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/main.dart
```diff
@@ -23,9 +28,42 @@ void main() async {
   await SecureStorage().clearOnReinstall();
   await dotenv.load(fileName: Assets.aApp);
 
+  final container = ProviderContainer();
+  final logApp = container.read(featureFlagsProvider.notifier).get(LoggerFeatureFlag.logApp);
+
+  if (logApp) {
+    Logger.init(verbose: true);
```
ice-endymion, line 35: If put it here, then I think we may miss some errors that may occur above

lib/main.dart
```diff
@@ -23,9 +28,42 @@ void main() async {
...
+  /// any errors that occur in isolate execution.
+  Isolate.current.addErrorListener(
+    RawReceivePort((List<dynamic> pair) async {
+      final errorAndStacktrace = pair;
+      Logger.error(
+        '[Isolate Error] ${errorAndStacktrace.first}',
+        stackTrace: errorAndStacktrace.last as StackTrace?,
+      );
+    }).sendPort,
+  );
```
ice-endymion, line 62: Do we need them in prod?

lib/main.dart
```diff
@@ -23,9 +28,42 @@ void main() async {
   await SecureStorage().clearOnReinstall();
   await dotenv.load(fileName: Assets.aApp);
 
+  final container = ProviderContainer();
```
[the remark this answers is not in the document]
ice-alcides: no, it runs before `initProvider`, because there is a used `ProviderContainer` and `UncontrolledProviderScope`

lib/main.dart
ice-alcides: no, it runs before initProvider, because there is a used ProviderContainer and UncontrolledProviderScope

lib/main.dart
ice-alcides: Ideally, we should integrate a service like `Crashlytics` or `Sentry`, but in this case, we'd need to replace the current Logger with another provider specifically for handling certain types of errors and exceptions. At the very least, we must disable `Talker/Logger` in production.

ice-alcides: I've took it from here - https://firebase.google.com/docs/crashlytics/customize-crash-reports?platform=flutter

lib/main.dart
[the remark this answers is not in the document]
ice-alcides: Flutter Hooks are designed for use within widget components (HookWidget or HookConsumerWidget), not at application startup, but I got your point and will create `LoggerInitializer` class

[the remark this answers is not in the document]
ice-alcides: done

ice-endymion (dismissed):

ice-endymion (approved):

lib/app/services/logger/logger_initializer.dart
```diff
@@ -0,0 +1,46 @@
...
+      RawReceivePort((List<dynamic> pair) async {
+        final errorAndStacktrace = pair;
+        Logger.error(
+          '[Isolate Error] ${errorAndStacktrace.first}',
+          stackTrace: errorAndStacktrace.last as StackTrace?,
+        );
+      }).sendPort,
+    );
+  }
+}
```
ice-kreios, line 46: we already have a `Logger` service,  instead of creating another logger initializer class should we do it in current one?

ice-alcides: It’s the same logger service, but `Logger/Talker` only outputs messages to the console without handling these types of errors.

ice-tychon (approved):

ice-kreios (approved):

PR #1006, opened 2025-04-10
fix: update user info menu on video scroll in trending videos

ice-alcides:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/video/views/pages/videos_vertical_scroll_page.dart
+++ b/lib/app/features/video/views/pages/videos_vertical_scroll_page.dart
@@ -103,6 +103,7 @@ class VideosVerticalScrollPage extends HookConsumerWidget {
             initialMediaIndex;
 
     final userPageController = usePageController(initialPage: initialPage);
+    final currentEventReference = useState<EventReference>(eventReference);
 
     return AnnotatedRegion<SystemUiOverlayStyle>(
       value: SystemUiOverlayStyle(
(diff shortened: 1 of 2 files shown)


lib/app/features/video/views/pages/videos_vertical_scroll_page.dart
```diff
@@ -139,7 +140,10 @@ class VideosVerticalScrollPage extends HookConsumerWidget {
           controller: userPageController,
           itemCount: flattenedVideos.length,
           scrollDirection: Axis.vertical,
-          onPageChanged: (index) => _loadMore(ref, index, flattenedVideos.length),
+          onPageChanged: (index) {
+            _loadMore(ref, index, flattenedVideos.length);
+            currentEventReference.value = flattenedVideos[index].entity.toEventReference();
```
ice-kreios, line 145: asking to be sure: If we set the event reference when the page changes, should we also check other functionalities? 

For example, in this widget, we retrieve the counts by 
`ref.watch(ionConnectEntityWithCountersProvider(eventReference: eventReference));`
it still uses the passed `eventReference` parameter

ice-endymion (approved):

lib/app/features/video/views/pages/videos_vertical_scroll_page.dart
ice-alcides: thanks, good catch, fixed

ice-kreios (approved):

PR #1016, opened 2025-04-11
fix: high memory consumption

ice-endymion:
## Description
Fix gallery media slow loading

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/services/media_service/media_service.c.dart
+++ b/lib/app/services/media_service/media_service.c.dart
@@ -48,8 +48,7 @@ class MediaFile with _$MediaFile {
 typedef CropImageUiSettings = List<PlatformUiSettings>;
 
 class MediaService {
-  final streamController = StreamController<List<MediaFile>>.broadcast();
-  List<MediaFile> _currentMediaFiles = [];
+  final _streamController = StreamController<List<MediaFile>>.broadcast();
   int _currentPage = 0;
   MediaPickerType? _currentType;
 
@@ -72,8 +71,7 @@ class MediaService {
 
       if (albums.isEmpty) {
         if (page == 0) {
-          _currentMediaFiles = [];
-          streamController.add([]);
+          _streamController.add([]);
         }
         return;
       }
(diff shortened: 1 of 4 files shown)


ice-alcides (dismissed):

ice-ajax (dismissed):

lib/app/services/media_service/media_service.c.dart
```diff
@@ -69,18 +69,35 @@ class MediaService {
...
+              ? asset.mimeType ?? await asset.mimeTypeAsync
+              : await asset.mimeTypeAsync;
+
+          mediaFiles.add(
+            MediaFile(
+              path: asset.id,
+              height: asset.height,
+              width: asset.width,
+              mimeType: mimeType,
+            ),
```
ice-kreios, line 94: If we will not run the method in parallel for each asset, why do we need batch implementation? To keep it simple, we can create media files synchronously, one by one. do you agree with me?

ice-endymion: resolved, add compute

lib/app/services/media_service/media_service.c.dart
```diff
@@ -69,18 +69,35 @@ class MediaService {
...
+      for (var i = 0; i < assets.length; i += batchSize) {
+        final end = (i + batchSize < assets.length) ? i + batchSize : assets.length;
+        final batch = assets.sublist(i, end);
+
+        // Process each asset in the batch directly with a for loop
+        for (final asset in batch) {
+          String? mimeType;
+
+          // For Android we can try to get the mimeType from the asset directly
+          mimeType = (defaultTargetPlatform == TargetPlatform.android)
```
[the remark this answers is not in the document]
ice-endymion: change the load process, so not actual anymore

ice-kreios (approved):

ice-kreios (approved):

PR #1039, opened 2025-04-17
feat: send funds for funds request in chat

ice-tychon:
## Description
This PR is a part of the Chat - Request Funds feature. It's responsible for showing a dialog to send funds as an answer to funds request in a chat.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots
CleanShot 2025-04-17 at 11 53 34

### DIFF
--- a/lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
+++ b/lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
@@ -15,8 +15,6 @@ import 'package:ion/app/features/wallets/providers/send_asset_form_provider.c.da
 import 'package:ion/app/features/wallets/providers/wallet_view_data_provider.c.dart';
 import 'package:ion/app/router/app_routes.c.dart';
 
-final _defaultMinimumSize = Size(150.0.s, 32.0.s);
-
 class MoneyMessageButton extends StatelessWidget {
   const MoneyMessageButton({
     required this.isMe,
@@ -33,14 +31,16 @@ class MoneyMessageButton extends StatelessWidget {
   final bool isPaid;
   final FundsRequestEntity? request;
 
+  static Size get _defaultMinimumSize => Size(150.0.s, 32.0.s);
+
   @override
   Widget build(BuildContext context) {
     if (isPaid) {
-      return const ViewTransactionButton();
+      return const _ViewTransactionButton();
     } else if (isMe && messageType == MoneyMessageType.requested) {
-      return CancelMoneyRequestButton(eventId: eventId);
+      return _CancelMoneyRequestButton(eventId: eventId);
     } else if (!isMe && messageType == MoneyMessageType.requested) {
-      return SendMoneyButton(eventId: eventId, request: request);
+      return _SendMoneyButton(eventId: eventId, request: request);
     } else {
       // Default case - shouldn't happen with proper conditions
       throw ArgumentError('No appropriate button for the given parameters');
@@ -48,10 +48,9 @@ class MoneyMessageButton extends StatelessWidget {
   }
 }
 
-class CancelMoneyRequestButton extends ConsumerWidget {
-  const CancelMoneyRequestButton({
+class _CancelMoneyRequestButton extends ConsumerWidget {
+  const _CancelMoneyRequestButton({
     required this.eventId,
-    super.key,
   });
 
   final String eventId;
@@ -61,7 +60,7 @@ class CancelMoneyRequestButton extends ConsumerWidget {
     return Button.compact(
       type: ButtonType.outlined,
       backgroundColor: context.theme.appColors.tertararyBackground,
-      minimumSize: _defaultMinimumSize,
+      minimumSize: MoneyMessageButton._defaultMinimumSize,
       label: Text(
         context.i18n.button_cancel,
         style: context.theme.appTextThemes.caption2.copyWith(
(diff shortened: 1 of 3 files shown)


lib/app/features/wallets/views/utils/network_validator.dart
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'dart:async';
+
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:ion/app/features/user/providers/user_metadata_provider.c.dart';
+import 'package:ion/app/features/wallets/model/network_data.c.dart';
+import 'package:ion/app/features/wallets/views/pages/contact_without_wallet_error_modal.dart';
+
+Future<bool> validateNetwork(
```
ice-brontes, line 10: I have concerns about this validator.

1. It seems strange that the network is considered valid if the `pubkey` is zero. I understand why you did it this way, but it seems to me that it is slightly inconsistent with the name of the method.
2. In fact, the method does not validate the network, but checks whether the selected user has a wallet in the selected network.

I would suggest changing the implementation slightly.
1. I would remake the method into a provider, for example, `doesUserHasWalletInNetwork` or something like that with the following implementation:
```
Future<bool> doesUserHasWalletInNetwork(Ref ref, String? pubkey, NetworkData network) async {
  final user = await ref.read(userMetadataProvider(pubkey).future);
  final walletsMap = user?.data.wallets ?? {};

  final hasProperWallet = walletsMap[network.id] != null;
  if (!hasProperWallet && ref.context.mounted) {
    unawaited(
      showContactWithoutWalletError(ref.context, user: user!, network: network),
    );
  }

  return hasProperWallet;
}
```
Perhaps we should add another flag that will tell whether to display the error dialog or not. Then the method will become more flexible, and we will be able to reuse it in different scenarios.

2. In the place where the current version is used, I would make the `_isNetworkValid` method with the following implementation:
```
Future<bool> _isNetworkValid(WidgetRef ref, NetworkData network, String? pubkey) {
  if (pubkey == null) return true;
  return ref.watch(doesUserHasWalletInNetworkProvider(pubkey, network).future);
}
```

WDYT?

lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
```diff
@@ -0,0 +1,188 @@
...
+      type: ButtonType.outlined,
+      backgroundColor: context.theme.appColors.darkBlue,
+      minimumSize: _defaultMinimumSize,
+      label: Text(
+        context.i18n.button_send,
+        style: context.theme.appTextThemes.caption2.copyWith(
+          color: context.theme.appColors.onPrimaryAccent,
+        ),
+      ),
+      onPressed: () async => _handleSendMoney(context, ref),
```
ice-kreios, line 102: wdyt handling in a provider and regarding that disabled the button while processing?

ice-kreios, line 102: It is a critical process, and we may need to display a confirmation modal to prevent unexpected taps while the user scrolls. What do you think?

Another suggestion: how to show error or success modals or banners? Sorry I missed it.

ice-tychon: The button doesn't send money right after it's pressed. There's a gif in the PR. You can see 2 dialogs (the form and the confirmation) after the click. The button just prepares the form to be shown in a prepared state.

ice-kreios (dismissed):

lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
ice-tychon: The second tap is gonna open the same page as the first one (if a user managed to tap it very fast). So nothing bad happens.

lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
```diff
@@ -0,0 +1,188 @@
...
+import 'package:ion/app/features/chat/model/money_message_type.dart';
+import 'package:ion/app/features/core/views/pages/error_modal.dart';
+import 'package:ion/app/features/wallets/model/entities/funds_request_entity.c.dart';
+import 'package:ion/app/features/wallets/providers/coins_provider.c.dart';
+import 'package:ion/app/features/wallets/providers/networks_provider.c.dart';
+import 'package:ion/app/features/wallets/providers/send_asset_form_provider.c.dart';
+import 'package:ion/app/features/wallets/providers/wallet_view_data_provider.c.dart';
+import 'package:ion/app/router/app_routes.c.dart';
+
+final _defaultMinimumSize = Size(150.0.s, 32.0.s);
```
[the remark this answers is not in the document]
ice-tychon: Fixed.

lib/app/features/chat/views/components/message_items/message_types/money_message/components/money_message_button.dart
```diff
@@ -0,0 +1,188 @@
...
+    } else if (!isMe && messageType == MoneyMessageType.requested) {
+      return SendMoneyButton(eventId: eventId, request: request);
+    } else {
+      // Default case - shouldn't happen with proper conditions
+      throw ArgumentError('No appropriate button for the given parameters');
+    }
+  }
+}
+
+class CancelMoneyRequestButton extends ConsumerWidget {
```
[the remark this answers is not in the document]
ice-tychon: Fixed.

lib/app/features/wallets/views/utils/network_validator.dart
ice-tychon: Good points. Fixed.

ice-ajax (approved):

PR #1054, opened 2025-04-21
feat: compressors enhancements

ice-endymion:
## Description
- Split `compress_service.c.dart` to dedicated 4 compressors:
    - `video_compressor.c.dart`
    - `audio_compressor.c.dart`
    - `image_compressor.c.dart`
    - `brotli_compressor.c.dart`
- Add presets for video and audio compressors

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/services/compressors/audio_compressor.c.dart
```diff
@@ -0,0 +1,109 @@
...
+import 'package:ion/app/services/compressors/compress_executor.c.dart';
+import 'package:ion/app/services/compressors/output_path_generator.dart';
+import 'package:ion/app/services/logger/logger.dart';
+import 'package:ion/app/services/media_service/ffmpeg_commands_config.dart';
+import 'package:ion/app/services/media_service/media_service.c.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'audio_compressor.c.g.dart';
+
+class AudioCompressorConfig {
```
ice-kreios, line 15:
```
enum AudioCompressorConfig {
  mediumQuality(bitrate: 96, sampleRate: 24000);

  const AudioCompressorConfig({
    required this.bitrate,
    required this.sampleRate,
  });

  final int bitrate;
  final int sampleRate;
}
```

wdyt, instead of using class/static fields, could we create an enhanced enum?

ice-endymion: Yes, sounds good

lib/app/features/chat/e2ee/providers/send_chat_message/compress_chat_media_provider.c.dart
```diff
@@ -14,16 +17,14 @@ part 'compress_chat_media_provider.c.g.dart';
 Raw<Future<MediaFile>> compressChatMedia(Ref ref, MediaFile mediaFile) async {
   final mediaType = MediaType.fromMimeType(mediaFile.mimeType ?? '');
 
-  final compressService = ref.watch(compressServiceProvider);
-
   switch (mediaType) {
     case MediaType.image:
-      return compressService.compressImage(mediaFile, shouldCompressGif: true);
+      return ref.watch(imageCompressorProvider).compressImage(mediaFile, shouldCompressGif: true);
```
ice-alcides, line 22: I think we could clean up the compression logic by extracting a `Compressor` interface and implementing each format as its own strategy class::

```
abstract interface class Compressor {
  Future<CompressedFile> compress(MediaFile file);
}
```
```
final class VideoCompressor implements Compressor
final class AudioCompressor implements Compressor
final class BrotliCompressor implements Compressor
```

factory provider
```
@riverpod
Compressor compressor(Ref ref, MediaType type) {
 return switch (type) {
      MediaType.image => ImageCompressor(),
      MediaType.video => VideoCompressor(),
      ...
}
```

Usage becomes:
```
final result = await ref
  .watch(compressorProvider(mediaType))
  .compress(mediaFile);
```

ice-endymion: I've been thinking about this, but what about parameterization for video, audio, image, brotli? `compress` func kinda different for each of them.

ice-endymion: Add common Compressor interface

lib/app/features/chat/e2ee/providers/send_chat_message/compress_chat_media_provider.c.dart
```diff
@@ -17,14 +13,13 @@ part 'compress_chat_media_provider.c.g.dart';
...
-    case MediaType.audio:
-      return ref.watch(audioCompressorProvider).compressAudio(mediaFile.path);
-    case MediaType.unknown:
-      return ref.watch(brotliCompressorProvider).compressWithBrotli(File(mediaFile.path));
+  final compressor = ref.watch(compressorProvider(mediaType));
+  if (mediaType == MediaType.unknown) {
+    return compressor.compress(
+      mediaFile,
+      settings: const ImageCompressionSettings(shouldCompressGif: true),
+    );
```
ice-kreios: Why do we need this block? Previously, if the media type is not image/video/audio (unknow), we were running the Brotli flow.

lib/app/features/chat/e2ee/providers/send_chat_message/compress_chat_media_provider.c.dart
```diff
@@ -17,14 +13,13 @@ part 'compress_chat_media_provider.c.g.dart';
...
   }
+
+  return compressor.compress(mediaFile);
```
ice-kreios, line 24: if it is an image, we should give `shouldCompressGif` as `true` because @ice-hector confirmed it

ice-alcides (dismissed):

lib/app/features/chat/e2ee/providers/send_chat_message/compress_chat_media_provider.c.dart
ice-endymion: same, typo, fixed.

ice-alcides (approved):

PR #1071, opened 2025-04-23
fix: stop video from pausing in full screen when mute is toggled

ice-ajax:
## Description
This PR fixes an issue where the video in full screen mode would pause when muted/unmuted

## Additional Notes
This issue is actually caused by the video_preview present inside feed cell underneath that uses the same video controller. It has a `onEffect` which pauses the video, because cell is not visible on the screen. By introducing a `wasOffScreen` flag we avoid video pausing after controller changes if the video_preview was already off screen before that.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/video/views/pages/video_page.dart
+++ b/lib/app/features/video/views/pages/video_page.dart
@@ -50,7 +50,7 @@ class VideoPage extends HookConsumerWidget {
               authorPubkey: authorPubkey,
               autoPlay: true,
               looping: looping,
-              uniqueId: framedEventReference?.encode() ?? '',
+              uniqueId: "${framedEventReference?.encode() ?? ''}-video-page",
             ),
           ),
         )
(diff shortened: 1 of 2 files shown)


ice-alcides (dismissed):

lib/app/features/video/views/pages/video_page.dart
```diff
@@ -50,7 +50,7 @@ class VideoPage extends HookConsumerWidget {
               authorPubkey: authorPubkey,
               autoPlay: true,
               looping: looping,
-              uniqueId: framedEventReference?.encode() ?? '',
+              uniqueId: "${framedEventReference?.encode() ?? ''}-video-page",
```
ice-kreios, line 53: If we do it, will the video play from the start even if the user has watched a little bit in the feed?

ice-endymion (dismissed):

ice-endymion (approved):

ice-alcides (approved):

ice-kreios (approved):

PR #1074, opened 2025-04-24
fix: prevent duplicate message issue when it is already sent for a masterpubkey

ice-kreios:
## Description
If a message has already been sent for a master pubkey, do not resend it to prevent duplicate messages.

## Additional Notes
N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


ice-kreios:
root cause will be  fixed by @ice-damocles, closing

PR #1088, opened 2025-04-25
fix: article description limits

ice-hector:
## Description
This PR fixes the article description char limits 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


ice-endymion (approved):

lib/app/features/feed/create_article/views/pages/article_form_modal/hooks/use_article_form.dart
```diff
@@ -68,7 +68,7 @@ ArticleFormState useArticleForm(WidgetRef ref, {EventReference? modifiedEvent})
   final descriptionOverflowCount = useState<int>(0);
   final isDescriptionLengthValid = useState(true);
   const titleMaxLength = 120;
-  const descriptionMaxLength = 25000;
+  const descriptionMaxLength = 25001;
```
ice-kreios, line 71: LGTM but I think we add an extra character limit of 1 because Quill adds a `/n` as the first character. Should we apply the same for the title max length too, wdyt?

ice-kreios (approved):

PR #1095, opened 2025-04-26
fix(wallet): show Wallet section only when it has data

ice-tychon:
## Description
Previously, the Friends section on the Wallet page was shown right after the friends' pubkeys are loaded. This lead to a situation when the section is shown, but friends are just Loaders.
This PR changes the way we show the Friends list. We show a Friends section loader until at least 1 of the friend's metadata is loaded (see video)

This same goes for Coins/Nfts section. It's gonna be shown only when it has data (see screenshot)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots
CleanShot 2025-04-26 at 19 16 18
image


lib/app/features/wallets/views/pages/wallet_page.dart
```diff
@@ -33,7 +35,12 @@ class WalletPage extends HookConsumerWidget {
 
     final activeTab = useState<WalletTabType>(WalletTabType.coins);
 
-    final showFriends = ref.watch(shouldShowFriendsSelectorProvider).valueOrNull ?? true;
+    final shouldShowLoader = ref.watch(shouldShowFriendsLoaderProvider);
+    final showFriendsSection = ref.watch(shouldShowFriendsSectionProvider);
+
+    final footer = SizedBox(
+      height: ScreenSideOffset.defaultSmallMargin,
+    );
```
ice-kreios, line 43: so simple question: why do we create the footer widget in the build method instead of directly passing 
```
FriendsListLoader(footer:  SizedBox(
      height: ScreenSideOffset.defaultSmallMargin,
    )
```

lib/app/features/wallets/views/pages/wallet_page/components/header/wallet_tabs_header_loader.dart
```diff
@@ -0,0 +1,23 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return ScreenSideOffset.small(
+      child: Column(
+        children: [
+          SizedBox(height: 8.0.s),
+          SkeletonBox(height: 40.0.s),
+          SizedBox(height: 2.0.s),
+        ],
```
ice-kreios, line 19: If we use the `Padding` widget, then we do not need the `Column` and two `SizedBox`. Does that make sense to you?

lib/app/features/wallets/views/pages/wallet_page.dart
ice-tychon: Fixed

ice-kreios (approved):

lib/app/features/wallets/views/pages/wallet_page/components/header/wallet_tabs_header_loader.dart
ice-tychon: Fixed.

ice-alcides (approved):

PR #1099, opened 2025-04-27
fix: notification tap zone

ice-hector:
## Description
This PR increases notification tappable area

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/notifications/views/notifications_history_page/components/notification_item/notification_item.dart
+++ b/lib/app/features/feed/notifications/views/notifications_history_page/components/notification_item/notification_item.dart
@@ -4,7 +4,6 @@ import 'package:flutter/material.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/components/screen_offset/screen_side_offset.dart';
 import 'package:ion/app/extensions/extensions.dart';
-import 'package:ion/app/features/feed/data/models/entities/article_data.c.dart';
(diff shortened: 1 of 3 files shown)


lib/app/features/feed/notifications/views/notifications_history_page/components/notification_item/notification_item.dart
```diff
@@ -70,4 +82,21 @@ class NotificationItem extends ConsumerWidget {
...
+    if (entity == null) {
+      return;
+    }
+
+    final eventReference = switch (entity) {
+      final GenericRepostEntity repost => repost.data.eventReference,
+      _ => entity.toEventReference(),
+    };
+
+    if (eventReference is ReplaceableEventReference && eventReference.kind == ArticleEntity.kind) {
```
ice-tychon, line 96: I don't like the idea of possibility of adding more `if-else if-else` into this method in future. Is it possible to refactor it to pass `onTap` into a Widget?

The bad thing is that we have a check `eventReference is ReplaceableEventReference && eventReference.kind == ArticleEntity.kind` in 2 places, here and in the `notification_related_entity.dart`.

ice-kreios (approved):

ice-tychon (dismissed):

ice-alcides (approved):

ice-kreios (approved):

PR #1101, opened 2025-04-27
fix: refresh followers count

ice-endymion:
## Description
Remove keep alive for `followersCount` provider, to get new state on profile page instead cached.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/providers/followers_count_provider.c.dart
+++ b/lib/app/features/user/providers/followers_count_provider.c.dart
@@ -11,7 +11,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'followers_count_provider.c.g.dart';
 
-@Riverpod(keepAlive: true)
+@riverpod
 FutureOr<int?> followersCount(
   Ref ref, {
   required String pubkey,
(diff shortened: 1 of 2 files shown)


ice-ajax (approved):

lib/app/features/user/providers/followers_count_provider.c.dart
```diff
@@ -11,7 +11,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'followers_count_provider.c.g.dart';
 
-@Riverpod(keepAlive: true)
+@riverpod
 FutureOr<int?> followersCount(
```
ice-ajax, line 15: This providers listens to the `countProvider`, which listens to cache, so it should rebuild, even when keepAlive, right?

ice-endymion: No, `followersCountProvider` just cached it state with `countProvider`, so `countProvider` called there build func only ones - when creating an instance.

lib/app/features/user/providers/followers_count_provider.c.dart
```diff
@@ -11,7 +11,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'followers_count_provider.c.g.dart';
 
-@Riverpod(keepAlive: true)
+@riverpod
```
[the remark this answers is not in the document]
ice-endymion: You right, we don't need it anymore, since we sync following and followers presentation. Removed.

ice-kreios (approved):

ice-ajax (approved):

PR #1125, opened 2025-04-30
feat(pushes): firebase messaging configuration

ice-orion:
## Description
This PR:
* Adds `DeviceIdService`
* Implements kind `31751` (`PushSubscription` event) and sends it depending to the app settings -> push notification options

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

image

### DIFF
--- a/lib/app/features/ion_connect/database/event_messages_database.c.steps.dart
+++ b/lib/app/features/ion_connect/database/event_messages_database.c.steps.dart
@@ -38,53 +38,38 @@ final class Schema2 extends i0.VersionedSchema {
 
 class Shape0 extends i0.VersionedTable {
   Shape0({required super.source, required super.alias}) : super.aliased();
-  i1.GeneratedColumn<String> get id =>
-      columnsByName['id']! as i1.GeneratedColumn<String>;
-  i1.GeneratedColumn<int> get kind =>
-      columnsByName['kind']! as i1.GeneratedColumn<int>;
-  i1.GeneratedColumn<String> get pubkey =>
-      columnsByName['pubkey']! as i1.GeneratedColumn<String>;
+  i1.GeneratedColumn<String> get id => columnsByName['id']! as i1.GeneratedColumn<String>;
+  i1.GeneratedColumn<int> get kind => columnsByName['kind']! as i1.GeneratedColumn<int>;
+  i1.GeneratedColumn<String> get pubkey => columnsByName['pubkey']! as i1.GeneratedColumn<String>;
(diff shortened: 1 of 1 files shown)


lib/app/services/device_id/device_id.c.dart
```diff
@@ -0,0 +1,85 @@
...
+class AndroidDeviceIdDelegate implements DeviceIdDelegate {
+  static const String _deviceIdKey = 'device_id';
+
+  @override
+  Future<String> getDeviceId() async {
+    final androidId = await const AndroidId().getId();
+    if (androidId != null) {
+      return androidId;
+    } else {
+      final sharedPrefs = await SharedPreferences.getInstance();
```
ice-alcides, line 52: can you replace with `sharedPreferencesProvider`? Then you can get the instance synchronously via`prefs.requireValue`

lib/app/services/device_id/device_id.c.dart
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+
+class IOSDeviceIdDelegate implements DeviceIdDelegate {
+  static const String _deviceIdKey = 'device_id';
+  static const String _deviceIdAccountName = 'device_id_account';
+
+  @override
+  Future<String> getDeviceId() async {
+    const storage = FlutterSecureStorage();
```
ice-alcides, line 70: can we use `secureStorageProvider`?

lib/app/features/push_notifications/providers/push_subscription_sync_provider.c.dart
```diff
@@ -0,0 +1,62 @@
...
+            selectedCategoriesSubscription,
+            actionSource: ActionSourceRelayUrl(selectedCategoriesSubscription.relay.url),
+          );
+    }
+  }
+
+  Future<void> _deleteSubscription(PushSubscriptionEntity entity) async {
+    await ref.read(ionConnectNotifierProvider.notifier).sendEntityData(
+          DeletionRequest(
+            events: [EventToDelete(eventId: entity.id, kind: PushSubscriptionEntity.kind)],
```
ice-kreios, line 56: As I see it, `PushSubscriptionEntity` is a replaceable event, but we delete it by `eventId`. I think we should send `a` tag instead of an `e` tag when the user wants to delete it. wdyt?

ice-alcides (approved):

ice-kreios (approved):

ice-endymion (approved):

PR #1132, opened 2025-05-01
feat: invalidate user metadata on profile pull to refresh

ice-ajax:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/pages/profile_page/profile_page.dart
+++ b/lib/app/features/user/pages/profile_page/profile_page.dart
@@ -6,8 +6,10 @@ import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/components/screen_offset/screen_top_offset.dart';
 import 'package:ion/app/components/separated/separator.dart';
 import 'package:ion/app/extensions/extensions.dart';
+import 'package:ion/app/features/ion_connect/providers/ion_connect_cache.c.dart';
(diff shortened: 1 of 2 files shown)


ice-alcides (approved):

lib/app/features/user/pages/profile_page/profile_page.dart
```diff
@@ -119,4 +128,9 @@ class ProfilePage extends HookConsumerWidget {
       ),
     );
   }
+
+  void _onRefresh(WidgetRef ref, UserMetadataEntity? userMetadata) {
+    if (userMetadata == null) return;
+    ref.read(ionConnectCacheProvider.notifier).remove(userMetadata.cacheKey);
```
ice-kreios, line 134: As I see it, if you remove it from cache, how does it trigger invalidation (re-fetch from source)?

ice-ajax: `userMetadataProvider` watches `ionConnectCacheProvider.notifier`. If metadata is removed from the cache, `userMetadataProvider` will try to rebuild, it will see that cache is empty so it will trigger a network request.

ice-alcides (approved):

ice-kreios (approved):

PR #1133, opened 2025-05-01
feat: add followers search

ice-endymion:
## Description
Add search for followers list
- Add `pagedFilter` to handle edge case with pagination

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/providers/followers_data_source_provider.c.dart
+++ b/lib/app/features/user/providers/followers_data_source_provider.c.dart
@@ -12,11 +12,17 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 part 'followers_data_source_provider.c.g.dart';
 
 @riverpod
-List<EntitiesDataSource>? followersDataSource(Ref ref, String pubkey) {
+List<EntitiesDataSource>? followersDataSource(
+  Ref ref,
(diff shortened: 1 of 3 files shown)


lib/app/features/user/providers/followers_data_source_provider.c.dart
```diff
@@ -12,11 +12,17 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
...
+  String pubkey, {
+  String? query,
+}) {
   return [
     EntitiesDataSource(
       actionSource: ActionSourceUser(pubkey),
-      entityFilter: (entity) => entity is FollowListEntity,
+      // Search not working for kind 3, so we use kind 0
+      entityFilter: (entity) =>
+          query != null ? entity is UserMetadataEntity : entity is FollowListEntity,
```
ice-endymion, line 25: @ice-orion Take a look on this one:
Search not work for kind 3, but work for kind 0 - so I set request filter with kind 0 only for search query, but then pagination not working. 
I can increase limit for search request as workaround, wdyt?

ice-alcides (approved):

PR #1139, opened 2025-05-03
refactor: extract FeedGlobalNotifications from GlobalNotifications

ice-tychon:
## Description
This PR makes Global Notifications more global. Previously they were tightly coupled to the feed.

### Changes
- Extracted `GlobalNotificationView` from `GlobalNotificationBar`
- `createPostNotifierProvider` listeners extracted from `GlobalNotificationBar` to a separate file in the Feed feature.
- Changed `GlobalNotificationNotifier` state type and fixed possible race condition with a `CancelableOperation`
- Introduced `GlobalNotification` type along with `FeedGlobalNotification` instead of the `GlobalNotificationData` to make it possible to reuse global notifications for other purposes, not only for Feed.
- Added `equatable` dependency (it already was in the project, but now it's direct, not transitive).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots

https://github.com/user-attachments/assets/4c40de09-3753-48ef-8073-2415a1bbf8b8


pubspec.lock
```diff
@@ -641,7 +641,7 @@ packages:
     source: hosted
     version: "3.0.0"
   equatable:
-    dependency: transitive
+    dependency: "direct main"
```
ice-kreios, line 644: Why didn’t you use a `freezed` package? If you don’t want it to generate fromJson, toJson, etc., extra auto-generated methods, we can disable or enable it like this? `@Freezed(copyWith: true)`

pubspec.yaml
```diff
@@ -32,6 +32,7 @@ dependencies:
   drift: ^2.24.0
   drift_flutter: ^0.2.4
   email_validator: ^3.0.0
+  equatable: ^2.0.7
```
ice-kreios, line 35: Why didn’t you use a freezed package? If you don’t want it to generate fromJson, toJson, etc., extra auto-generated methods, we can disable or enable it like this? @Freezed(copyWith: true)

ice-tychon: Fixed.

ice-kreios (approved):

lib/app/router/components/app_router_builder.dart
```diff
@@ -18,6 +19,7 @@ class AppRouterBuilder extends HookConsumerWidget {
     return Material(
       color: context.theme.appColors.secondaryBackground,
       child: GlobalNotificationBarWrapper(
+        setUpListeners: setupFeedGlobalNotificationsListeners,
```
ice-endymion, line 22: As I understand the idea was to split notifications by scopes, like feed, profile, chat, global, etc, so I think `setUpListeners` should handle List of listeners.

ice-endymion (approved):

ice-alcides (approved):

PR #1146, opened 2025-05-05
feat: add No Internet Connection global notification

ice-tychon:
## Description
This PR adds a No Internet Connection global notification.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots 
image


ice-alcides (approved):

ice-ajax (dismissed):

pubspec.yaml
```diff
@@ -22,6 +22,7 @@ dependencies:
   camera: ^0.11.0+2
   charcode: ^1.4.0
   collection: ^1.18.0
+  connectivity_plus: ^6.1.4
```
ice-kreios, line 25: https://pub.dev/packages/connectivity_plus

`You should not rely on the current connectivity status to decide whether you can reliably make a network request. Always guard your app code against timeouts and errors that might come from the network layer. Connection type availability does not guarantee that there is an Internet access. For example, the plugin might return Wi-Fi connection type, but it might be a connection with no Internet access due to network requirements (like on hotel Wi-Fi networks where user often needs to go through a captive portal to authorize first).`

Are we sure we are aware of this warning from the packages' documentation? 

We may also check [internet_connection_checker_plus](https://pub.dev/packages/internet_connection_checker_plus)

ice-tychon: The PR uses the package you suggested. I just left the `connectivity_plus` as a direct dependecy. I'll remove it.

ice-ajax (approved):

ice-alcides (approved):

PR #1160, opened 2025-05-07
fix: ion transactions balance

ice-brontes:
## Description
1. Removed unused methods from the `TransactionsRepository`.
2. Fixed order of the transactions on the coin details page.
3. Add `externalHash` to the transaction data model.
4. Add external hash processor to handle transactions with this field.
5. `SyncTransactionsService` broken into smaller pieces without changing functionality. 

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/wallets/domain/transactions/transfer_status_updater.c.dart
```diff
@@ -0,0 +1,76 @@
...
+  final TransactionsRepository _transactionsRepository;
+
+  Future<void> update(Wallet wallet) async {
+    final broadcasted = await _transactionsRepository.getBroadcastedTransfers(
+      walletAddress: wallet.address,
+    );
+    if (broadcasted.isEmpty) return;
+
+    final updated = <TransactionData>[];
+    String? nextPageToken = '';
```
ice-endymion, line 36: does it make sense to init with empty string? why not just leave with null?

ice-brontes: Cos `while` will work only with non-nullable `nextPageToken`. It's like the flag, that we haven't more pages to load, so we need to stop syncing.

ice-kreios (approved):

ice-tychon (approved):

ice-endymion (approved):

PR #1180, opened 2025-05-10
fix: skip display of address not found modal for ion network

ice-ajax:
## Description
This PR adds a logic where `AddresNotFoundModal` is not displayed for ION network and the wallet address is immediately created.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
+++ b/lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
@@ -2,9 +2,12 @@
 
 import 'package:flutter/foundation.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:ion/app/features/components/verify_identity/verify_identity_prompt_dialog_helper.dart';
+import 'package:ion/app/features/wallets/model/network_data.c.dart';
 import 'package:ion/app/features/wallets/views/pages/coins_flow/receive_coins/providers/receive_coins_form_provider.c.dart';
 import 'package:ion/app/features/wallets/views/pages/coins_flow/receive_coins/providers/wallet_address_notifier_provider.c.dart';
 import 'package:ion/app/hooks/use_on_init.dart';
+import 'package:ion_identity_client/ion_identity.dart';
 
 void useCheckWalletAddressAvailable(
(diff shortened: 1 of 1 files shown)


lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
```diff
@@ -18,10 +21,44 @@ void useCheckWalletAddressAvailable(
 
       if (address != null) {
         ref.read(receiveCoinsFormControllerProvider.notifier).setWalletAddress(address);
+      } else if (address == null &&
+          network != null &&
+          network.displayName == 'ION' &&
+          ref.context.mounted) {
+        await _createIonWallet(ref, network);
```
ice-kreios, line 28: A small comment: I think we do not need to await it because not doing anything after that will not be beneficial. wdyt?

lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
```diff
@@ -18,10 +21,44 @@ void useCheckWalletAddressAvailable(
 
       if (address != null) {
         ref.read(receiveCoinsFormControllerProvider.notifier).setWalletAddress(address);
+      } else if (address == null &&
+          network != null &&
+          network.displayName == 'ION' &&
+          ref.context.mounted) {
```
ice-kreios, line 27: we may check `ref.context.mounted` at the top of the entire if clause to avoid writing it for each `else if` clause. wdyt?

lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
ice-ajax: removed await

lib/app/features/wallets/hooks/use_check_wallet_address_available.dart
ice-ajax: fixed

ice-endymion (approved):

ice-kreios (approved):

PR #1185, opened 2025-05-11
feat(android): push translations match

ice-orion:
## Description
This PR introduces data push notification handling on android and passing required data to iOS Notification Extension 

## Additional Notes
* Add sharing of `current_master_pubkey` and `app_locale` with the notification service extension via app group feature
* Update shared prefs lib, migrate to a new api, update related tests
* Add `IonConnectGiftWrapEntity`
* Add generic `Translator` class for handling translations with versions and cache, use it for notification translations
* Add `IonConnectPushDataPayload` that is responsible for parsing / processing data push payload
* Add `notificationResponseData` provider (will be used in next PR)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/push_notifications/providers/app_translations_provider.c.dart
+++ b/lib/app/features/push_notifications/providers/app_translations_provider.c.dart
@@ -179,8 +179,7 @@ class AppTranslationsRepository<T extends TranslationWithVersion> {
     return '${directory.path}/${_getCacheFileName(locale: locale)}';
   }
 
-  String _getCacheFileName({required Locale locale}) =>
-      'app_translations.${locale.languageCode}.json';
+  String _getCacheFileName({required Locale locale}) => '$_path.${locale.languageCode}.json';
 
   String _getCacheVersionKey({required Locale locale}) => 'cache_version_${locale.languageCode}';
 }


ice-kreios (approved):

lib/app/services/storage/local_storage.c.dart
```diff
@@ -19,44 +43,51 @@ LocalStorage localStorage(Ref ref) {
   return LocalStorage(prefs.requireValue);
 }
 
+@Riverpod(keepAlive: true)
+Future<LocalStorage> localStorageAsync(Ref ref) async {
```
ice-alcides, line 47: Why do you need another provider `localStorageAsyncProvider`? We already initialized `sharedPreferencesProvider` in `initProvider`, so in my opinion, it would be enough to use `localStorageProvider`

ice-alcides: I guess it's because we call `initProvider` in the wrong place. It should be called earlier, as when the app starts using the `container.read` (you can look at `main.dart` for logger initialization). In this case, all providers from initAppProvider would be correctly eagerly initialized

ice-alcides (approved):

PR #1198, opened 2025-05-13
fix: Set same createdAt time

ice-damocles:
## Description
Fixed issue when messages are marked as "edited"

## Type of Change
- [x] Bug fix

### DIFF
--- a/lib/app/features/chat/e2ee/model/entities/private_direct_message_data.c.dart
+++ b/lib/app/features/chat/e2ee/model/entities/private_direct_message_data.c.dart
@@ -132,11 +132,9 @@ class ReplaceablePrivateDirectMessageData
     DateTime? createdAt,
     DateTime? publishedAtTime,
   }) {
-    final createdAt = DateTime.now();
-
     return EventMessage.fromData(
       signer: signer,
-      createdAt: createdAt,
+      createdAt: createdAt ?? DateTime.now(),
       kind: ReplaceablePrivateDirectMessageEntity.kind,
       content: content,
       tags: [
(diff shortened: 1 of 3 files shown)


ice-alcides (approved):

ice-tychon (approved):

lib/app/features/chat/e2ee/model/entities/private_direct_message_data.c.dart
```diff
@@ -132,11 +132,9 @@ class ReplaceablePrivateDirectMessageData
     DateTime? createdAt,
     DateTime? publishedAtTime,
   }) {
-    final createdAt = DateTime.now();
-
     return EventMessage.fromData(
       signer: signer,
-      createdAt: createdAt,
+      createdAt: createdAt ?? DateTime.now(),
```
ice-kreios, line 137:
```
static FutureOr<EventMessage> fromData({
    required EventSigner signer,
    required int kind,
    required String content,
    List<List<String>> tags = const [],
    DateTime? createdAt,
  }) async {
    createdAt ??= DateTime.now();
    final normalizedTags = _normalizeTags(tags);
```

It will do the same if `createdAt` is `null`. Do we need it? Otherwise, LGTM.

ice-kreios (approved):

lib/app/features/chat/e2ee/model/entities/private_direct_message_data.c.dart
ice-damocles: But seems like it is equal approach `createdAt ?? DateTime.now()` or `createdAt ??= DateTime.now()`

PR #1234, opened 2025-05-17
fix: make link preview in post body clickable

ice-ajax:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/views/components/url_preview_content/url_preview_content.dart
+++ b/lib/app/features/feed/views/components/url_preview_content/url_preview_content.dart
@@ -5,6 +5,7 @@ import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:ion/app/components/image/ion_network_image.dart';
 import 'package:ion/app/components/url_preview/url_preview.dart';
 import 'package:ion/app/extensions/extensions.dart';
+import 'package:ion/app/services/browser/browser.dart';
 import 'package:ion/app/utils/url.dart';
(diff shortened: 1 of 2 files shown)


ice-alcides (approved):

lib/app/features/feed/views/components/url_preview_content/url_preview_content.dart
```diff
@@ -13,38 +14,44 @@ part 'components/url_metadata_preview.dart';
...
-      url: url,
-      builder: (meta, favIconUrl) {
-        if (meta == null || meta.title == null || meta.title!.isEmpty) {
-          return const SizedBox.shrink();
-        }
+    return GestureDetector(
+      onTap: clickable ? () => openUrlInAppBrowser(normalizeUrl(url)) : null,
+      behavior: HitTestBehavior.opaque,
+      child: UrlPreview(
+        key: ValueKey(url),
```
ice-kreios, line 30: If there are two posts with the same URL in the feed, I suppose it will throw a duplicate key error while rendering it on the widget tree. wdyt?

ice-ajax: Probably not, because these are local keys on different branches of the render tree. It would happen if we used GlobalKeys here

ice-endymion (approved):

ice-kreios (approved):

PR #1251, opened 2025-05-20
feat: add generic typed config repository with cache and version support

ice-kreios:
## Description
- Introduced generic ConfigRepository supporting typed config retrieval via parser function
- Implemented caching logic with support for both local storage and file-based cache
- Enabled optional version-based fetching for AppConfigWithVersion implementations
- Added ConfigParser<T> abstraction to simplify parser and metadata handling

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/config/providers/config_repository.c.dart
```diff
@@ -0,0 +1,176 @@
...
+    required LocalStorage localStorage,
+  })  : _dio = dio,
+        _ionOrigin = ionOrigin,
+        _localStorage = localStorage;
+
+  final Dio _dio;
+  final String _ionOrigin;
+  final LocalStorage _localStorage;
+
+  Future<T> getConfig<T>(
```
ice-tychon, line 33: There's no protection against multiple simultaneous requests for the same config, which could cause race conditions and duplicate network calls.

lib/app/features/config/providers/config_repository.c.dart
```diff
@@ -0,0 +1,176 @@
...
+      if (response.statusCode == HttpStatus.noContent) {
+        return null;
+      } else {
+        return parser(
+          response.data is String ? response.data as String : jsonEncode(response.data),
+        );
+      }
+    } catch (error) {
+      Logger.error(error);
+      throw AppConfigException(error);
```
ice-tychon, line 83: Consider adding some additional message to the `AppConfigException`, like `throw AppConfigException(message, error)`, where message could be `failed to get from network` or `failed to get from cache`. It could make it easier to find the issue later.

ice-alcides (approved):

ice-tychon (approved):

lib/app/features/config/providers/config_repository.c.dart
ice-kreios: I implemented a lock mechanism using `config_name`, allowing the first call to retrieve data from the network, which is then saved into cache. Subsequent calls will access this data from the cache.

lib/app/features/config/providers/config_repository.c.dart
ice-kreios: fixed

ice-tychon (approved):

PR #1301, opened 2025-05-27
fix: enable swipe-down to dismiss fullscreen page on Android

ice-kreios:
## Description
This PR fix enables the swipe-down gesture to dismiss the fullscreen page on Android. Previously, this feature was not functional when users swiped down to close the page.

## Additional Notes
N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/router/components/modal_wrapper/dismissible_page.dart
+++ b/lib/app/router/components/modal_wrapper/dismissible_page.dart
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: ice License 1.0
 
-import 'package:flutter/gestures.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:ion/app/utils/future.dart';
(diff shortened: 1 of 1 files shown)


lib/app/router/components/modal_wrapper/dismissible_page.dart
```diff
@@ -130,39 +129,17 @@ class DismissiblePage extends HookWidget {
...
-          },
-        ),
-      };
-    } else {
-      gestures = {};
-    }
-
-    return RawGestureDetector(
-      gestures: gestures,
+    return GestureDetector(
```
ice-alcides, line 132: can you check please if this affected the double tap zoom in a post with multiple images?

ice-alcides: Using the usual `GestureDetector` instead of `RawGestureDetector` can cause race conditions with zooming and attempts to swiped down to close the page

ice-alcides (approved):

PR #1344, opened 2025-06-03
fix: notification sorting order

ice-hector:
## Description
This PR fixes the notifications sorting order

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/notifications/data/repository/followers_repository.c.dart
+++ b/lib/app/features/feed/notifications/data/repository/followers_repository.c.dart
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: ice License 1.0
 
 import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:ion/app/extensions/extensions.dart';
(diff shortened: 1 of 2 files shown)


lib/app/features/feed/notifications/data/repository/followers_repository.c.dart
```diff
@@ -36,7 +36,9 @@ class FollowersRepository implements IonNotificationRepository {
         .map(
           (result) => FollowersIonNotification(
             total: result.uniquePubkeyCount,
-            timestamp: DateTime.parse(result.eventDate),
+            timestamp: DateTime.fromMillisecondsSinceEpoch(
+              (result.lastCreatedAt ?? 0) * 1000,
+            ),
```
ice-ajax, line 41: It's safer to call `result.lastCreatedAt.toDateTime` since recently we changed time precision from seconds to microseconds, so `lastCreatedAt` could be either of those

ice-ajax (approved):

PR #1377, opened 2025-06-05
feat: change the way we save and load encrypted keypair file

ice-tychon:
## Description
Previously we were renaming ION Identity key (from `/keys`) to `device-{file_id}`, where `file_id` – is the id of the file we uploaded to Relays. We were doing that to be able to find the file later.

Now, we want to send `imeta` tag to `kind 0` (current user metadata) with the information of the file, to be able to find the file on another device. This PR makes these changes.

## Type of Change
- [x] New feature
- [x] Breaking change


ice-ajax (approved):

lib/app/features/ion_connect/providers/upload_device_keypair_notifier.c.dart
```diff
@@ -132,14 +131,16 @@ class UploadDeviceKeypairNotifier extends _$UploadDeviceKeypairNotifier {
...
-      height: 1,
       path: tempFile.path,
       mimeType: 'application/octet-stream',
     );
 
     try {
-      return await ref.read(ionConnectUploadNotifierProvider.notifier).upload(mediaFile);
+      return await ref.read(ionConnectUploadNotifierProvider.notifier).upload(
+            mediaFile,
+            alt: FileAlt.attestationKey,
```
ice-kreios, line 141: as far as I understand, @ice-ares suggested we should use `alt` only for visual media, and we should use the `summary` prop for the device key pair. otherwise LGTM

ice-kreios (approved):

PR #1393, opened 2025-06-06
feat: add maintenance page

ice-ajax:
## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

ScreenShot 2025-06-06 at 11 27 23@2x


ice-alcides (approved):

lib/app/features/core/views/pages/error_page.dart
ice-kreios, line 1: So, when an uncaught error happens in the app, we show an error page. But let's think about it from the user's perspective. Are we sure we should display a maintenance page? Maybe the server is up, but another error occurred. I think we should only show it when the user can't reach the relays. wdyt? otherwise lgtm

ice-kreios (approved):

lib/app/features/core/views/pages/error_page.dart
ice-ajax: confirmed it should be shown in case any error happens during app launch

ice-alcides (approved):

PR #1465, opened 2025-06-18
feat: user metadata from database

ice-damocles:
## Description
- Syncs E2EE conversations user metadata to the local database
- Pull to refresh (always forces sync of existing conversation profiles)
- New event (30014, 5, 7) triggers check of the users metadata (sync window controller by env variable)
- User opened one-to-one conversation triggers check of users metadata
- Any new master pubkey passed to the flows which doesn't exist in the db causes force update

## Type of Change
- [x] New feature
- [x] Refactoring

### DIFF
--- a/lib/app/features/chat/e2ee/views/pages/one_to_one_messages_page.dart
+++ b/lib/app/features/chat/e2ee/views/pages/one_to_one_messages_page.dart
@@ -19,7 +19,7 @@ import 'package:ion/app/features/chat/recent_chats/providers/selected_reply_mess
 import 'package:ion/app/features/chat/views/components/message_items/edit_message_info/edit_message_info.dart';
 import 'package:ion/app/features/chat/views/components/message_items/messaging_bottom_bar/messaging_bottom_bar.dart';
 import 'package:ion/app/features/chat/views/components/message_items/replied_message_info/replied_message_info.dart';
-import 'package:ion/app/features/user/providers/user_metadata_provider.c.dart';
+import 'package:ion/app/features/user_metadata/providers/user_metadata_from_db_provider.c.dart';
 import 'package:ion/app/router/app_routes.c.dart';
 import 'package:ion/app/services/media_service/media_service.c.dart';
 import 'package:ion/app/utils/username.dart';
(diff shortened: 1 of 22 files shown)


lib/app/features/chat/e2ee/views/pages/one_to_one_messages_page.dart
```diff
@@ -110,22 +122,22 @@ class _Header extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final receiverUserMetadata = ref.watch(userMetadataProvider(receiverMasterPubkey));
+    final receiverUserMetadata =
+        ref.watch(userMetadataFromDbNotifierProvider(receiverMasterPubkey));
 
-    if (receiverUserMetadata.isLoading) {
+    if (receiverUserMetadata == null) {
```
ice-kreios, line 128: If the receiver is deleted, then there won't be any metadata. However, we need to show the messages. Can you check this case, please?

lib/app/features/user_metadata/database/user_metadata_database.c.dart
```diff
@@ -0,0 +1,44 @@
...
+import 'package:ion/app/features/ion_connect/database/converters/event_reference_converter.c.dart';
+import 'package:ion/app/features/ion_connect/database/converters/event_tags_converter.dart';
+import 'package:ion/app/features/ion_connect/model/event_reference.c.dart';
+import 'package:ion/app/features/user_metadata/database/tables/user_metadata_table.c.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'user_metadata_database.c.g.dart';
+
+@riverpod
+UserMetadataDatabase userMetadataDatabase(Ref ref) {
```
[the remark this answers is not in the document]
ice-damocles: Initial requirement was to have separate database cause it also will be accessed by ios push notification background service

lib/app/features/chat/e2ee/views/pages/one_to_one_messages_page.dart
ice-damocles: This is just conversation header, I added skeleton loader if there is no metadata so user will still be able to see messages etc

PR #1473, opened 2025-06-19
feat: allow unclassified posts in feed fetch

ice-hades:
## Description
- Allow unclassified posts in feed fetch;

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/data/models/feed_modifier.dart
+++ b/lib/app/features/feed/data/models/feed_modifier.dart
@@ -5,12 +5,14 @@ import 'package:ion/app/features/feed/data/models/feed_interests.c.dart';
 import 'package:ion/app/features/ion_connect/model/related_hashtag.c.dart';
 import 'package:ion/app/features/ion_connect/model/search_extension.dart';
 
+typedef FeedFilterData = ({List<SearchExtension> search, Map<String, List<String>> tags});
+
 enum FeedModifier {
   top,
   trending,
   explore;
 
-  ({List<SearchExtension> search, Map<String, List<String>> tags}) filter(FeedConfig config) {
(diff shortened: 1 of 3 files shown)


lib/app/features/feed/data/models/feed_modifier.dart
```diff
@@ -9,7 +10,7 @@ enum FeedModifier {
   trending,
   explore;
 
-  ({List<SearchExtension> search, Map<String, List<String>> tags}) get filter {
+  ({List<SearchExtension> search, Map<String, List<String>> tags}) filter(FeedConfig config) {
```
ice-tychon, line 13: Extract a typedef or a class for the return value.

Also, passing the entire `FeedConfig` object to the `filter` method violates the SOL`I`D. The method only needs `excludeUnclassifiedFromExplore` but receives a large config object with many unrelated fields.

Consider extracting only the required parameter:
```dart
ReturnType filter({
  bool excludeUnclassifiedFromExplore = false,
}) {
}
```

Or even create a FilterStrategy to make it easier to add and modify behaviour (maybe an overkill in this situation).

ice-tychon: I mean 
```
typedef FeedFilter = ({List<SearchExtension> search, Map<String, List<String>> tags})

// then
FeedFilter filter(...) { ... }
```
to improve readability.

> if we just require excludeUnclassifiedFromExplore meaning any value can be passed
Then create a class which retrieves `FeedConfig` and creates `FeedFilter` without requiring `excludeUnclassifiedFromExplore`. Basically, the factory I mentioned previously.

ice-kreios (approved):

PR #1475, opened 2025-06-20
feat: account notifications

ice-hector:
## Description
This PR implements subscription to different accounts in the app notifications types

## Task ID
ION-2920

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Screenshot 2025-06-20 at 12 32 48
Screenshot 2025-06-20 at 12 32 53

### DIFF
--- a/lib/app/features/user/providers/account_notifications_sync_provider.r.dart
+++ b/lib/app/features/user/providers/account_notifications_sync_provider.r.dart
@@ -32,54 +32,8 @@ class AccountNotificationsSync extends _$AccountNotificationsSync {
 
   @override
   FutureOr<void> build() async {
-    await initializeSync();
+    await _initializeSync();
     ref.onDispose(cancelAllSync);
-  }
-
-  Future<void> initializeSync() async {
-    final syncInterval = getSyncInterval();
-
-    final lastSyncTime = await getLastSyncTime();
-
-    if (shouldSyncImmediately(lastSyncTime, syncInterval)) {
-      await syncAndScheduleNext(syncInterval);
-    } else {
-      scheduleDelayedSync(lastSyncTime!, syncInterval);
-    }
-  }
-
-  Duration getSyncInterval() {
-    return ref.read(envProvider.notifier).get<Duration>(
-          EnvVariable.ACCOUNT_NOTIFICATION_SETTINGS_SYNC_INTERVAL_MINUTES,
-        );
-  }
-
-  bool shouldSyncImmediately(DateTime? lastSyncTime, Duration syncInterval) {
-    if (lastSyncTime == null) return true;
-
-    final timeSinceLastSync = DateTime.now().difference(lastSyncTime);
-    return timeSinceLastSync >= syncInterval;
-  }
-
-  void scheduleDelayedSync(DateTime lastSyncTime, Duration syncInterval) {
-    final timeSinceLastSync = DateTime.now().difference(lastSyncTime);
-    final remainingTime = syncInterval - timeSinceLastSync;
-
-    _syncTimer = Timer(remainingTime, () async {
-      await syncAndScheduleNext(syncInterval);
-    });
-  }
-
-  Future<void> syncAndScheduleNext(Duration syncInterval) async {
-    await performSync();
-    setupPeriodicSync(syncInterval);
-  }
-
-  void setupPeriodicSync(Duration interval) {
-    _syncTimer?.cancel();
-    _syncTimer = Timer.periodic(interval, (_) async {
-      await performSync();
-    });
   }
 
   void cancelAllSync() {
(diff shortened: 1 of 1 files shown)


lib/app/features/core/providers/env_provider.c.dart
```diff
@@ -82,6 +83,8 @@ class Env extends _$Env {
       EnvVariable.BLOCKED_USERS_SYNC_OVERLAP_DAYS =>
         const int.fromEnvironment('BLOCKED_USERS_SYNC_OVERLAP_DAYS') as T,
       EnvVariable.CHECKSUM => const String.fromEnvironment('CHECKSUM') as T,
+      EnvVariable.ACCOUNT_NOTIFICATION_SETTINGS_SYNC_INTERVAL_MINUTES =>
+        const int.fromEnvironment('ACCOUNT_NOTIFICATION_SETTINGS_SYNC_INTERVAL_MINUTES') as T,
```
ice-ajax, line 87: Might be worth adding `defaultValue` here, otherwise it is `0`

lib/app/features/user/providers/account_notifications_sets_provider.c.dart
```diff
@@ -0,0 +1,258 @@
...
+  Ref ref,
+  String currentPubkey,
+  List<UserNotificationsType> enabledNotifications,
+  Map<String, List<String>> currentSets,
+) async {
+  final setTypesToUsers = <String, List<String>>{
+    'in_app_notifications_posts': [],
+    'in_app_notifications_stories': [],
+    'in_app_notifications_articles': [],
+    'in_app_notifications_videos': [],
```
ice-ajax, line 137: could we extract this keys to some consts? I see they are duplicated like 4 times across this file.

lib/app/features/user/providers/account_notifications_sync_provider.c.dart
```diff
@@ -0,0 +1,458 @@
...
+    int? since,
+    int? until,
+  }) {
+    if (users.isEmpty) {
+      return null;
+    }
+
+    return switch (contentType) {
+      NotificationContentType.videos => RequestFilter(
+          kinds: const [1, 6, 30175, 1630175],
```
ice-ajax, line 339: could we reuse the kinds from the domain objects themselves, to that it is more readable? so `PostEntity.kind` instead of `1` etc...

lib/app/features/user/providers/account_notifications_sync_provider.c.dart
```diff
@@ -0,0 +1,458 @@
...
+  IonConnectEntity? _convertEventToEntity(EventMessage event) {
+    try {
+      return switch (event.kind) {
+        1 => PostEntity.fromEventMessage(event),
+        6 => RepostEntity.fromEventMessage(event),
+        7 => ReactionEntity.fromEventMessage(event),
+        30023 => ArticleEntity.fromEventMessage(event),
+        30175 => ModifiablePostEntity.fromEventMessage(event),
+        1630023 => GenericRepostEntity.fromEventMessage(event),
+        1630175 => GenericRepostEntity.fromEventMessage(event),
```
ice-ajax, line 400: same here about the kinds

lib/app/features/user/providers/user_specific_notifications_provider.c.dart
```diff
@@ -0,0 +1,70 @@
...
+          } else {
+            await removeUserFromNotificationSet(ref, userPubkey, type);
+          }
+        });
+        await Future.wait(updateFutures);
+
+        state = AsyncValue.data(selectedTypes);
+      }
+
+      ref.invalidateSelf();
```
ice-ajax, line 64: is the invalidate doing anything in this situation? We are watching `usersForNotificationTypeProvider` inside the build method, so if it changes this provider will rebuild automatically, but if it didn't change, then this invalidate won't do anything.

lib/app/features/feed/notifications/data/repository/content_repository.r.dart
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  ContentType? _getContentType(IonConnectEntity entity) {
+    return switch (entity) {
+      PostEntity() when entity.data.videos.isNotEmpty => ContentType.videos,
+      PostEntity() when entity.data.expiration != null => ContentType.stories,
+      PostEntity() => ContentType.posts,
+      ModifiablePostEntity() when entity.data.videos.isNotEmpty => ContentType.videos,
+      ModifiablePostEntity() when entity.data.expiration != null => ContentType.stories,
+      ModifiablePostEntity() => ContentType.posts,
```
ice-kreios, line 70: First, we should check if the enttiy has an `expiration`. If not, if the user shares a story with a video media, it will be processed as `ContentType.videos` instead of `ContentType.stories`.

lib/app/features/user/providers/account_notifications_sync_provider.r.dart
```diff
@@ -0,0 +1,385 @@
...
+import 'package:ion/app/features/user/model/account_notifications_sets.f.dart';
+import 'package:ion/app/features/user/model/user_notifications_type.dart';
+import 'package:ion/app/features/user/providers/user_relays_manager.r.dart';
+import 'package:ion/app/utils/algorithm.dart';
+import 'package:riverpod_annotation/riverpod_annotation.dart';
+
+part 'account_notifications_sync_provider.r.g.dart';
+
+@Riverpod(keepAlive: true)
+class AccountNotificationsSync extends _$AccountNotificationsSync {
```
ice-kreios, line 30: `keepAliveWhenAuthenticated` can we use it to dispose when user logs out?

lib/app/features/user/providers/account_notifications_sync_provider.r.dart
```diff
@@ -0,0 +1,385 @@
...
+      onEvent: (event) {
+        eventFutures.add(_processNotificationEvent(event, contentType));
+      },
+      actionSource: ActionSourceRelayUrl(relayUrl),
+    );
+
+    await Future.wait(eventFutures);
+
+    final actualNewTimestamp = newLastCreatedAt == latestEventTimestamp.microsecondsSinceEpoch
+        ? latestEventTimestamp.add(const Duration(microseconds: 1))
```
ice-kreios, line 288: Why do you need to add 1 millisecond? I believe we should always save the timestamp of the latest received event.

PR #1555, opened 2025-07-04
 fix: auto-update profile feed when creating posts

ice-alcides:
## Description
When creating a post from the user profile, the new post was not automatically displayed in the profile feed

## Task ID
3138

## Type of Change
- [х] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/pages/profile_page/components/tabs/tab_entities_list.dart
+++ b/lib/app/features/user/pages/profile_page/components/tabs/tab_entities_list.dart
@@ -7,6 +7,7 @@ import 'package:ion/app/components/scroll_view/pull_to_refresh_builder.dart';
 import 'package:ion/app/features/auth/providers/auth_provider.m.dart';
 import 'package:ion/app/features/components/entities_list/entities_list.dart';
 import 'package:ion/app/features/components/entities_list/entities_list_skeleton.dart';
+import 'package:ion/app/features/feed/providers/user_posts_provider.r.dart';
 import 'package:ion/app/features/ion_connect/model/ion_connect_entity.dart';
 import 'package:ion/app/features/ion_connect/providers/entities_paged_data_provider.m.dart';
 import 'package:ion/app/features/user/model/tab_entity_type.dart';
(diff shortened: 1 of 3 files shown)


lib/app/features/user/pages/profile_page/components/tabs/tab_entities_list.dart
```diff
@@ -54,15 +55,23 @@ class TabEntitiesList extends ConsumerWidget {
...
+      onLoadMore: () async {
+        if (type == TabEntityType.posts) {
+          await ref.read(userPostsProvider(pubkey).notifier).fetchEntities();
+        } else {
+          final dataSource = ref.read(tabDataSourceProvider(type: type, pubkey: pubkey));
+          if (dataSource != null) {
+            await ref.read(entitiesPagedDataProvider(dataSource).notifier).fetchEntities();
+          }
+        }
+      },
```
ice-kreios, line 73: Do we need to use async/await? We might delete async/await because we don’t use the returned data.wdyt?

lib/app/features/user/pages/profile_page/components/tabs/tab_entities_list.dart
```diff
@@ -95,7 +104,14 @@ class TabEntitiesList extends ConsumerWidget {
...
         onRefresh: () async {
-          ref.invalidate(entitiesPagedDataProvider(dataSource));
+          if (type == TabEntityType.posts) {
+            ref.invalidate(userPostsProvider(pubkey));
+          } else {
+            final dataSource = ref.read(tabDataSourceProvider(type: type, pubkey: pubkey));
+            if (dataSource != null) {
+              ref.invalidate(entitiesPagedDataProvider(dataSource));
+            }
+          }
```
ice-kreios, line 114: same

lib/app/features/user/pages/profile_page/components/tabs/tab_entities_list.dart
ice-alcides: 1. `LoadMoreBuilder.onLoadMore` expects `Future<void> Function()`
2. `PullToRefreshBuilder.onRefresh` expects `Future<void> Function()`

PR #1561, opened 2025-07-04
feat: feed following requests limit

ice-orion:
## Description
This PR implements the limitation of max requests to fetch data from the following users. We score the users and prioritize those most likely to have new content.

## Additional Notes
* Fix the issue with DB `FeedModifierConverter`. `FeedModifier` is a part of the composite primary key in some tables, so it should never be nullable on DB level. 

## Task ID
ION-3145

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/data/database/following_feed_database/dao/user_fetch_states_dao.m.dart
+++ b/lib/app/features/feed/data/database/following_feed_database/dao/user_fetch_states_dao.m.dart
@@ -55,8 +55,8 @@ class UserFetchStatesDao extends DatabaseAccessor<FollowingFeedDatabase>
       ..where(
         // `FeedModifierConverter` should be used manually here,
         // because otherwise there is a false positive issue that it throws an exception that we're trying
-        // to pass a null value to a nun-nullable column, even though `FeedModifierConverter().toSql`
-        // returns nun-nullable int.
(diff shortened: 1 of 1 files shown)


lib/app/features/feed/data/database/following_feed_database/dao/seen_events_dao.m.dart
```diff
@@ -31,9 +41,7 @@ class SeenEventsDao extends DatabaseAccessor<FollowingFeedDatabase> with _$SeenE
       ..where((table) => table.eventReference.equalsValue(eventReference))
       ..where((table) => table.feedType.equalsValue(feedType))
       ..where(
-        (table) => feedModifier == null
-            ? table.feedModifier.isNull()
-            : table.feedModifier.equalsValue(feedModifier),
+        (tbl) => tbl.feedModifier.equals(const FeedModifierConverter().toSql(feedModifier)),
```
ice-kreios, line 44: `(table) =>  table.feedModifier.equalsValue(feedModifier)`

Why don’t we check it like this?

ice-kreios (approved):

PR #1563, opened 2025-07-07
fix: prevent loop when no previous interest data

ice-kreios:
## Description
This PR prevents unnecessary recursive calls in `_fetchInterestsEntities` by ensuring we only recurse when `requestedCount > 0`. Previously, the function could call itself even when no new entities were returned, potentially causing infinite or redundant loops when no data is available from the relays.

## Additional Notes
N/A

## Task ID
3159-3160-3163

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


ice-alcides (approved):

ice-kreios:
https://github.com/ice-blockchain/ion-app/pull/1566
closing it because root cause fixed with https://github.com/ice-blockchain/ion-app/pull/1566

PR #1570, opened 2025-07-08
feat: improve config fetch

ice-endymion:
## Description
- Add version check for `ConfigRepositoryProvider`
- Add fallback version from X-Version response header
- Fix rare bug when app with old config model gets latest config version without new keys

## Task ID
3102

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/config/providers/config_repository.r.dart
+++ b/lib/app/features/config/providers/config_repository.r.dart
@@ -107,22 +107,23 @@ class ConfigRepository {
     AppConfigCacheStrategy cacheStrategy, {
     int? version,
   }) async {
-    final (data, headerVersion) = await _getFromNetwork<T>(
+    final result = await _getFromNetwork<T>(
       configName,
       parser,
       checkVersion,
       version: version,
     );
 
-    if (data != null) {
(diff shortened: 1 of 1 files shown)


ice-kreios (approved):

lib/app/features/config/providers/config_repository.r.dart
```diff
@@ -159,10 +205,25 @@ class ConfigRepository {
...
+  void _verifyNetworkData<T>(String configName, T data, bool checkVersion, int? version) {
+    if (checkVersion && data is! AppConfigWithVersion && version == null) {
+      final error = Exception(
+        'Config "$configName" must implement AppConfigWithVersion or include version information in the response header, when checkVersion is true.',
+      );
+      Logger.error(error);
+      throw AppConfigException(
+        error,
+        configName: configName,
+        errorMessage: 'Invalid config version type',
```
ice-alcides, line 217: should we localize this string?

PR #1591, opened 2025-07-09
fix(chat): show all user search results

ice-kreios:
## Description
Resolves an issue where chat user search results remained in a loading state. The implementation now falls back to fetching user metadata from `cachedUserMetadataProvider` when the data is not available in the local database.

## Additional Notes
N/A

## Task ID
3203

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/search/model/chat_search_result_item.f.dart
```diff
@@ -0,0 +1,14 @@
...
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+part 'chat_search_result_item.f.freezed.dart';
+
+@freezed
+class ChatSearchResultItem with _$ChatSearchResultItem {
+  const factory ChatSearchResultItem({
+    required String masterPubkey,
+    required String lastMessageContent,
+    @Default(false) bool isFromLocalDb,
```
ice-alcides, line 12: Embedding `isFromLocalDb` in ChatSearchResultItem leaks persistence details into a DTO and tightly couples layers. Could we move the source‑resolution logic into the provider/repository instead, so the model remains storage‑agnostic?

ice-kreios: good point! fixed

PR #1598, opened 2025-07-10
fix: do not filter out current user events, handle reposts corner case

ice-orion:
## Description
This PR fixes the issue on feed when the app filters out all events from the current user, that leads to an issue when we receive our event, filter returns null, we stop querying the data source assuming that we don't have events left. 

## Task ID
ION-3213

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/providers/feed_data_source_builders.dart
+++ b/lib/app/features/feed/providers/feed_data_source_builders.dart
@@ -1,7 +1,5 @@
 // SPDX-License-Identifier: ice License 1.0
 
-import 'package:collection/collection.dart';
-import 'package:ion/app/exceptions/exceptions.dart';
 import 'package:ion/app/extensions/extensions.dart';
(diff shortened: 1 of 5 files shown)


ice-alcides (approved):

lib/app/features/feed/providers/feed_data_source_builders.dart
```diff
@@ -91,10 +88,6 @@ EntitiesDataSource buildVideosDataSource({
   return EntitiesDataSource(
     actionSource: actionSource,
     entityFilter: (entity) {
-      if (entity.masterPubkey == currentPubkey) {
-        return false;
-      }
```
ice-kreios, line 96: As far as I recall, it was added to prevent circular dependency errors on the feed. If this no longer occurs, LGTM!

ice-kreios (approved):

PR #1622, opened 2025-07-15
fix: prevent flicker when rendering URL metadata in feed scroll

ice-kreios:
## Description
Fixes a flickering issue in the feed caused by posts with URLs triggering metadata rendering during scroll.

## Additional Notes
N/A

## Task ID
3267

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/eb3bea10-b31f-4ace-afd9-83a98d6ca5b5

### DIFF
--- a/lib/app/features/feed/views/components/post/components/post_body/post_body.dart
+++ b/lib/app/features/feed/views/components/post/components/post_body/post_body.dart
@@ -90,15 +90,9 @@ class PostBody extends HookConsumerWidget {
           return false;
         }
 
-        if (content.isSingleLinkOnly) {
-          if (hasValidUrlMetadata) {
-            if (media.isEmpty) {
-              return false;
-            }
-            return true;
-          }
+        if (content.isSingleLinkOnly && hasValidUrlMetadata && media.isEmpty) {
+          return false;
         }
-
         return true;
       },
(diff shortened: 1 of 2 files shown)


ice-ajax (approved):

lib/app/features/feed/views/components/post/components/post_body/post_body.dart
```diff
@@ -62,31 +64,44 @@ class PostBody extends HookConsumerWidget {
...
+        if (content.isSingleLinkOnly) {
+          if (hasValidUrlMetadata) {
+            if (media.isEmpty) {
+              return false;
+            }
+            return true;
+          }
+        }
+
+        return true;
```
ice-ajax, line 102: Could it be simplified to this?

```suggestion
        if (content.isSingleLinkOnly && hasValidUrlMetadata && media.isEmpty) {
           return false;
        }
        return true;
```

ice-kreios: you are right, i just fixed

ice-ajax (approved):

PR #1635, opened 2025-07-16
fix: share posts to chat in parallel

ice-damocles:
## Description
- Send posts to the participants in parallel

## Task ID
ION-3266

## Type of Change
- [x] Bug fix

### DIFF
--- a/lib/app/features/chat/providers/share_feed_item_to_chat_provider.r.dart
+++ b/lib/app/features/chat/providers/share_feed_item_to_chat_provider.r.dart
@@ -64,114 +64,153 @@ class ShareFeedItemToChat extends _$ShareFeedItemToChat {
 
         await Future.wait(
           receiversMasterPubkeys.map((masterPubkey) async {
-            final existingConversationId =
-                await ref.read(existChatConversationIdProvider(masterPubkey).future);
-
-            final conversationId = existingConversationId ??
-                sendChatMessageService.generateConversationId(receiverPubkey: masterPubkey);
-
-            final tags = [
-              MasterPubkeyTag(value: currentUserMasterPubkey).toTag(),
-              ['k', feedItemEventMessage.kind.toString()],
-              [RelatedPubkey.tagName, eventSigner.publicKey],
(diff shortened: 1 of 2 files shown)


lib/app/features/chat/providers/share_feed_item_to_chat_provider.r.dart
```diff
@@ -62,106 +62,114 @@ class ShareFeedItemToChat extends _$ShareFeedItemToChat {
...
-                  pubkey: pubkey,
-                  eventSigner: eventSigner,
-                  eventMessage: kind16Rumor,
-                  masterPubkey: masterPubkey,
-                  wrappedKinds: [
-                    GenericRepostEntity.kind.toString(),
-                    feedItemEventMessage.kind.toString(),
-                  ],
-                );
+        await Future.wait(
```
ice-alcides, line 65: The current code has three nested `Future.wait` blocks. That makes the flow hard to follow and can launch a lot of simultaneous operations at once. Could we refactor so we:

- move all logic for one receiver into a helper
- call a single Future.wait over the list of helpers
- keep the inner loops as simple for statements?

ice-damocles: Okay, I moved code to two methods `sendToReceiver` and `sendKind16ToUserDevice` as they correspond to the business, but will keep inner `Future.wait` cause thats exactly part needed to be refactored in order to share post to multiple users

ice-alcides: If the receiver list can grow, the fan-out becomes `receivers × 2 × devices`, which could flood the DB/network - is that the case? If yes, I propose wrapping the top‑level loop with a simple concurrency limiter using the Pool package (https://pub.dev/packages/pool)

ice-alcides (approved):

lib/app/features/chat/providers/share_feed_item_to_chat_provider.r.dart
ice-alcides: discussed

PR #1646, opened 2025-07-17
fix: same profile push

ice-hector:
## Description
This PR fixes navigation to the same profile from the profile screen

## Task ID
ION-3269

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/router/utils/profile_navigation_utils.dart
+++ b/lib/app/router/utils/profile_navigation_utils.dart
@@ -12,8 +12,7 @@ class ProfileNavigationUtils {
     final currentState = GoRouterState.of(context);
     final currentPath = currentState.matchedLocation;
 
-    final isCurrentlyOnProfile =
-        currentPath.startsWith('/profile/user/') || currentPath.contains('/user/');
+    final isCurrentlyOnProfile = currentPath.contains('/user/');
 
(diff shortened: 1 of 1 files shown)


ice-tychon (dismissed):

ice-alcides (dismissed):

lib/app/router/utils/profile_navigation_utils.dart
```diff
@@ -0,0 +1,27 @@
...
+class ProfileNavigationUtils {
+  static Future<T?> navigateToProfile<T extends Object?>(
+    BuildContext context,
+    String pubkey,
+  ) {
+    final currentState = GoRouterState.of(context);
+    final currentPath = currentState.matchedLocation;
+
+    final isCurrentlyOnProfile =
+        currentPath.startsWith('/profile/user/') || currentPath.contains('/user/');
```
ice-kreios, line 16: The second condition `currentPath.contains('/user/')` already covers `currentPath.startsWith('/profile/user/')`, so the first check is redundant. do you agree with me?

ice-alcides (approved):

ice-tychon (approved):

PR #1672, opened 2025-07-20
fix: video page gradient + missing shadows and buttons

ice-cerberus:
## Description
- changed paddings and sizes on video page footer
- fixed a case when the video text content containing only just a line break was considered non empty
- added missing kebab button
- adjusted mute button
- added shadows to video post info text, kebab button, and avatar

## Task ID
ION-3286

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Before:
image

Aftrer:
image

image

### DIFF
--- a/lib/app/features/feed/views/components/user_info/user_info.dart
+++ b/lib/app/features/feed/views/components/user_info/user_info.dart
@@ -20,6 +20,7 @@ class UserInfo extends HookConsumerWidget {
     this.createdAt,
     this.accentTheme = false,
     this.timeFormat = TimestampFormat.short,
+    this.shadow,
     super.key,
   });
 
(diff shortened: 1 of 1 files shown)


lib/app/features/video/views/components/video_post_text.dart
```diff
@@ -52,13 +52,14 @@ class VideoTextPost extends HookWidget {
       ],
     );
 
-    if (content.isEmpty) {
+    if (content.isEmpty || content.first.data.toString().trim().isEmpty) {
```
ice-kreios, line 55: super small suggestion: we already have a [pre defined logic](https://github.com/ice-blockchain/ion-app/blob/master/lib/app/components/text_editor/text_editor_preview.dart#L66`) to handle it but it is not reusable, we might create an extension on Delta class to reuse it. wdyt?

ice-kreios (approved):

ice-kreios (approved):

PR #1691, opened 2025-07-22
fix: link attribute regex pattern

ice-hades:
## Description
- Fixed link attribute regex pattern to recognize links like ice.io and at the time ice.but is not a valid link;
- Fixed quill editor issue when you type a valid link like ice.io and a link attribute is already applied to it then you continue typing and ice.ion is not a valid link already so you end up with ice.io part marked as link and n is not but actually link attribute should be removed from ice.io part

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/72c54c72-2ae7-4040-8183-566a3b685282

### DIFF
--- a/lib/app/services/text_parser/model/text_matcher.dart
+++ b/lib/app/services/text_parser/model/text_matcher.dart
@@ -24,11 +24,25 @@ class UrlMatcher extends TextMatcher {
   const UrlMatcher();
 
   @override
-  String get pattern => r'\b(?:(?:[a-z][a-z0-9+\-.]*):\/\/|www\.)' // scheme:// or www.
+  String get pattern => r'\b(?:'
+      // scheme or www with host (fallback allowed)
+      '(?:'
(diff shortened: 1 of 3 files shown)


lib/app/services/text_parser/model/text_matcher.dart
```diff
@@ -24,11 +24,24 @@ class UrlMatcher extends TextMatcher {
   const UrlMatcher();
 
   @override
-  String get pattern => r'\b(?:(?:[a-z][a-z0-9+\-.]*):\/\/|www\.)' // scheme:// or www.
+  String get pattern => r'\b(?:'
```
ice-kreios, line 27: `https://github.com/ice-blockchain/ion-app/issues?q=test`
it is an edge case but i just tested and it does not support a link with a query parameter. should we handle it too?

Simulator Screenshot - iPhone 16 Pro - 2025-07-22 at 20 17 32

PR #1692, opened 2025-07-22
feat: crypto formatting

ice-brontes:
## Description
Apply advanced rules to crypto formatting.

## Additional Notes
Rules:
**0 < value < 0.000001 (10 * (10^-7))** => display the value in the next way `0.0(n)d`, where
n - number of zeros after the dot
d - first non-zero value
**0.000001 < value < 10** => display max 6 decimal digits + integer part 
**value > 10** => display only 2 decimal digits + full integer part

## Task ID
ION-2515

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Simulator Screenshot - iPhone 16 Pro - 2025-07-22 at 17 10 37

### DIFF
--- a/lib/app/features/wallets/views/utils/crypto_formatter.dart
+++ b/lib/app/features/wallets/views/utils/crypto_formatter.dart
@@ -9,8 +9,11 @@ String formatCrypto(double value, [String? currency]) {
     // For values 1-9.99: max 6 decimals, min 2 decimals
     // Handle truncation for cases like 1.1234567 -> 1.123456 and 2.0000001 -> 2.00
     _ when value >= 1 => _formatWithSmartTruncation(value, maxDecimals: 6, minDecimals: 2),
-    // For values < 1: max 6 decimals, min 2 decimals, but handle very small numbers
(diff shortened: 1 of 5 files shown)


ice-tychon (approved):

ice-alcides (approved):

lib/app/features/wallets/views/utils/crypto_formatter.dart
```diff
@@ -3,9 +3,95 @@
...
+  final stringValue = value.toString();
+
+  if (!stringValue.contains('.')) {
+    // It's an integer, just format with minimum decimals
+    return formatDouble(
+      value,
+      maximumFractionDigits: maxDecimals,
+      minimumFractionDigits: minDecimals,
+    );
+  }
```
ice-kreios, line 39: i think we should take care about i18n, because decimal seperator might be different depend on the current user locale. however for now we format it with en_US locale. wdyt?

PR #1709, opened 2025-07-24
fix: account notifications resync

ice-hector:
## Description
This PR fixed the account notifications resync issue

## Task ID
ION-3140

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/providers/account_notifications_sync_provider.r.dart
+++ b/lib/app/features/user/providers/account_notifications_sync_provider.r.dart
@@ -38,7 +38,7 @@ class AccountNotificationsSync extends _$AccountNotificationsSync {
       return;
     }
 
-    final currentPubkey = ref.read(currentPubkeySelectorProvider);
+    final currentPubkey = ref.watch(currentPubkeySelectorProvider);
     if (currentPubkey == null) {
       return;
     }
@@ -63,7 +63,7 @@ class AccountNotificationsSync extends _$AccountNotificationsSync {
       }
     } catch (error) {
       // Fallback: schedule a retry
-      Timer(syncInterval, () async {
+      _syncTimer = Timer(syncInterval, () async {
         await _initializeSync();
       });
     }


ice-alcides (dismissed):

ice-damocles (approved):

lib/app/features/user/providers/account_notifications_sync_provider.r.dart
```diff
@@ -32,6 +32,17 @@ class AccountNotificationsSync extends _$AccountNotificationsSync {
   @override
   FutureOr<void> build() async {
     keepAliveWhenAuthenticated(ref);
+
+    final authState = await ref.watch(authProvider.future);
+    if (!authState.isAuthenticated) {
+      return;
+    }
+
+    final currentPubkey = ref.read(currentPubkeySelectorProvider);
```
ice-kreios, line 41: i think it should be `watch` instead of `read`, wdyt?

lib/app/features/user/providers/account_notifications_sync_provider.r.dart
```diff
@@ -43,13 +54,18 @@ class AccountNotificationsSync extends _$AccountNotificationsSync {
...
+    try {
+      final lastSyncTime = await _getLastSyncTime();
+      if (_shouldSyncImmediately(lastSyncTime, syncInterval)) {
+        await _syncAndScheduleNext(syncInterval);
+      } else {
+        _scheduleDelayedSync(lastSyncTime!, syncInterval);
+      }
+    } catch (error) {
+      // Fallback: schedule a retry
+      Timer(syncInterval, () async {
```
ice-kreios, line 66: I think we should set a `Timer` reference and cancel it when this provider is disposed. similar to the exists `_syncTimer`? Do you agree with me?

ice-alcides (approved):

PR #1718, opened 2025-07-25
fix: only latest video on feed should play

ice-hades:
## Description
- Only latest video on feed should play feature;

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/core/providers/video_player_provider.r.dart
+++ b/lib/app/features/core/providers/video_player_provider.r.dart
@@ -74,12 +74,17 @@ class VideoController extends _$VideoController {
           .select((state) => state[params.sourcePath] ?? params.sourcePath),
     );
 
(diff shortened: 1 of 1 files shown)


lib/app/features/core/providers/video_player_provider.r.dart
```diff
@@ -141,6 +145,18 @@ class VideoController extends _$VideoController {
             );
           }
         });
+
+        if (params.onlyOneShouldPlay) {
+          controller.addListener(() {
```
ice-kreios, line 150: i am just corious about should we remove listener when controller ll be disposed?

ice-kreios (approved):

PR #1719, opened 2025-07-26
feat: watch nft transactions to hide them after sending (until the broadcasted status will be updated to confirmed)

ice-brontes:
## Description
Remove transferred NFTs from the wallet view.

## Additional Notes
This PR slightly improves/changes the wallet view managing.
Here is the updated components:

`WalletViewsService`
1. Fetches the wallet views (`_originWalletViews`); 
2. In the `_updateEmittedWalletViews` we apply filtering (remove from the wallet view all transferred NFTs, adjust transferred coins from the balances);
3. Next, a subscription to the `WalletViewsLiveUpdater` is made (`_liveUpdater.watchWalletViews`). 

`WalletViewsLiveUpdater`
Since we now need to monitor a lot of things, I moved all the logic for modifying the wallet view and subscribing to updates to a separate class. This class is responsible for applying filters + watching the coin prices + NFT/coins transfers. Whenever the new update is come: 
1. It takes the last updates from other streams (coin prices, NFT broadcasted transfers, coin broadcasted transfers).
2. Updates the coin balances, removes transferred NFTs, updates the coin prices.
3. Emits the updated wallet views. 

`WalletViewsDataNotifier`
Also, this class was a bit changed. Now it tracks the broadcasted transfers and if the number of them changes for some wallet view, it updates the wallet view to get the latest data from the server. It prevents us from the next issue:
1. User send NFT.
2. We hide this NFT from the wallet view.
3. During periodic sync of transaction, this transfer receives the `Confirmed` status. Now, we shouldn't hide this NFT manually.
4. Live updater emits the new state for wallet views, but since we haven't updated origin wallet views without sent NFT, we show it again.

`TransactionsDao`
Small guard was added here. Since we support NFT transactions now, I assume, that we can try to parse a transaction, which will not be identified as NFT or coin transaction. For that purpose, `_mapRowToDomainModel` can return null and log message about failure. 

## Task ID
ION-3210

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


ice-kreios (dismissed):

ice-alcides (dismissed):

lib/app/features/wallets/data/database/dao/transactions_dao.m.dart
```diff
@@ -125,14 +130,16 @@ class TransactionsDao extends DatabaseAccessor<WalletsDatabase> with _$Transacti
...
         transactionCoinAlias: transactionCoinAlias,
+        assetType: assetType,
       ),
       limit: limit,
       offset: offset,
-    ).watch().map((rows) => rows.toList());
+    ).watch().map((transactions) => transactions.whereType<TransactionData>().toList());
   }
 
   Future<List<TransactionData>> getTransactions({
```
ice-tychon, line 140: Can we split Coins and NFTs transactions somehow? The most of the code is the same, but adding coin ids and nft ids next to each other looks strange.

lib/app/features/wallets/data/repository/transactions_repository.m.dart
```diff
@@ -352,66 +378,70 @@ class TransactionsRepository {
...
-            alternatives: transaction.froms,
-            fallbackAddress: type.isSend ? wallet.address : null,
-          );
-          final to = _resolveTransactionAddress(
-            direct: transaction.to,
-            alternatives: transaction.tos,
-            fallbackAddress: !type.isSend ? wallet.address : null,
-          );
+          // If coin lookup failed, try to parse as NFT transaction
+          if (transaction.tokenId != null && contract != null) {
```
ice-tychon, line 409: Also here, the method's name is `loadCoinTransactions`, but it loads NFTs too.

ice-brontes: You are right, renamed method.

ice-alcides (approved):

ice-tychon (approved):

PR #1733, opened 2025-07-28
fix: context menu not appear for create post

ice-hades:
## Description
- Fixed context  copy/paste menu for create post not appear while no text is types yet. Figured it is due to non empty placeholder string. So some quill quirk. If placeholder string is non empty on null - issues with content menu. So I used a custom placeholder text using Stack

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/3c25b81a-4305-43ab-b6cb-b2e096601bed

### DIFF
--- a/lib/app/features/feed/create_post/views/pages/post_form_modal/components/create_post_content.dart
+++ b/lib/app/features/feed/create_post/views/pages/post_form_modal/components/create_post_content.dart
@@ -185,11 +185,30 @@ class _TextInputSection extends HookConsumerWidget {
                 mainAxisSize: MainAxisSize.min,
                 crossAxisAlignment: CrossAxisAlignment.start,
                 children: [
-                  TextEditor(
-                    textEditorController,
-                    placeholder: createOption.getPlaceholder(context),
-                    key: textEditorKey,
-                    scrollController: scrollController,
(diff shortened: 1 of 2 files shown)


ice-tychon (approved):

lib/app/features/feed/create_post/views/pages/post_form_modal/components/create_post_content.dart
```diff
@@ -185,11 +185,30 @@ class _TextInputSection extends HookConsumerWidget {
...
                 children: [
-                  TextEditor(
-                    textEditorController,
-                    placeholder: createOption.getPlaceholder(context),
-                    key: textEditorKey,
-                    scrollController: scrollController,
+                  AnimatedBuilder(
+                    animation: textEditorController,
+                    builder: (context, _) {
+                      final isEmpty = textEditorController.document.toPlainText().trim().isEmpty;
```
ice-kreios, line 191: We already have an `isBlank` extension for Delta, so we might use that here instead. wdyt?

ice-kreios (approved):

PR #1751, opened 2025-07-30
fix: fallback to en locale if config not found

ice-hades:
## Description
- Fallback for en locale if 404 for other locales when fetching `content-topics_${type}_$locale` remote configs;
-  Added padding at the bottom for language selection page when bottom bar is rendered cause it hides bottom of the languages list;
- Removed other option from languages list;

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Chore

### DIFF
--- a/lib/app/features/core/views/pages/language_selector_page.dart
+++ b/lib/app/features/core/views/pages/language_selector_page.dart
@@ -13,7 +13,7 @@ import 'package:ion/app/features/auth/views/pages/select_languages/language_list
 import 'package:ion/app/features/core/model/language.dart';
 import 'package:ion/app/features/core/providers/app_locale_provider.r.dart';
 import 'package:ion/app/hooks/use_languages.dart';
-import 'package:ion/app/hooks/use_on_init.dart';
+import 'package:ion/app/hooks/use_measured_widget_height.dart';
 import 'package:ion/app/router/components/navigation_app_bar/collapsing_app_bar.dart';
 import 'package:ion/app/router/components/navigation_app_bar/navigation_app_bar.dart';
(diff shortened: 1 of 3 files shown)


lib/app/features/feed/providers/feed_user_interests_provider.r.dart
```diff
@@ -74,13 +76,29 @@ class FeedUserInterests extends _$FeedUserInterests {
...
+    }
+
+    try {
+      return await fetchFor(locale);
+    } on AppConfigException catch (e) {
+      if (e.originalError is DioException &&
+          (e.originalError as DioException).response?.statusCode == 404 &&
+          locale != 'en') {
+        // Retry with en locale on 404
+        return fetchFor('en');
```
antonybrro, line 98: 💅 move 'en' to variable and use in condition and as parameter for `fetchFor`

lib/app/features/core/views/pages/language_selector_page.dart
```diff
@@ -44,10 +46,24 @@ class LanguageSelectorPage extends HookConsumerWidget {
     final languages = useLanguages(query: searchQuery.value, preferredLangs: preferredLangs);
 
     final mayContinue = selectedLanguages.isNotEmpty;
+    final renderBottomBar = mayContinue && continueButton != null;
+
+    final bottomBarKey = useMemoized(GlobalKey.new);
```
ice-tychon, line 51: Should it be extract to a new hook?

ice-tychon (approved):

PR #1752, opened 2025-07-30
fix: remove image precache from the CoinsGroupItem

ice-brontes:
## Description
`CoinsGroupItem` is used on different pages, like `Receive coins`, `Send coins` modals, etc. These pages have a search. For example, a user would paste the word "ice" in the search and see 100 different coins with the word "ice" in the name, symbol, or something else.

Because of the line
`ref.read(networkSelectorNotifierProvider(symbolGroup: coinsGroup.symbolGroup));`
for each of these 100 coins
`ref.watch(syncedCoinsBySymbolGroupProvider(symbolGroup));`
would be triggered, which would result in the mobile client making hundreds (or thousands) of requests in one session.

## Task ID
ION-3460

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/wallets/views/components/coins_list/coin_item.dart
+++ b/lib/app/features/wallets/views/components/coins_list/coin_item.dart
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: ice License 1.0
 
 import 'package:flutter/material.dart';
-import 'package:flutter_hooks/flutter_hooks.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/components/coins/coin_icon.dart';
 import 'package:ion/app/components/list_item/list_item.dart';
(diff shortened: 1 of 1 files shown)


lib/app/features/wallets/views/components/coins_list/coin_item.dart
```diff
@@ -32,19 +29,6 @@ class CoinsGroupItem extends HookConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final isBalanceVisible = ref.watch(isBalanceVisibleSelectorProvider);
 
-    final icons = ref.read(networkSelectorNotifierProvider(symbolGroup: coinsGroup.symbolGroup));
-
-    useEffect(
-      () {
-        if (icons != null) {
-          precachePictures(icons.items.map((item) => item.image));
```
ice-kreios, line 40: i totally agree, but what about the original issue? as far as i remember there were a icon rendering issue so @ice-neoptolemus added caching flow. we might cache it inside the `CoinIconWidget` with `_precachePicture`?

ice-tychon (approved):

PR #1777, opened 2025-08-01
fix: ui fixes

ice-erebus:
## Description
- Fixed ui related issues with post and vote
- Fixed vote option keys cause in cases where we have same options (e.g. "no" and "no") we got same keys and it throws exception

## Task ID
ION-3506 (UI related part)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Screenshot_1754055994

### DIFF
--- a/lib/app/features/feed/polls/view/components/poll_vote.dart
+++ b/lib/app/features/feed/polls/view/components/poll_vote.dart
@@ -25,7 +25,7 @@ class PollVote extends HookConsumerWidget {
         return List.generate(
           pollData.options.length,
           (index) => PollVoteItem(
-            key: ValueKey('poll_option_${pollData.options[index]}'),
+            key: ValueKey('poll_option_${pollData.options[index]}_$index'),
             text: pollData.options[index],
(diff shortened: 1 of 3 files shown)


lib/app/features/feed/polls/view/components/poll_vote.dart
```diff
@@ -25,7 +25,7 @@ class PollVote extends HookConsumerWidget {
         return List.generate(
           pollData.options.length,
           (index) => PollVoteItem(
-            key: ValueKey('poll_option_${pollData.options[index]}'),
+            key: ValueKey('poll_option_${pollData.options[index]}_$index'),
```
ice-kreios, line 28: it is super small edge case but lets assume user created 2 posts with same poll with same options and both of them visible on the screen. then it ll cause same the issue right? so i would suggest passing the `postReference` to make it unique from the parent widget (PostPoll). wdyt? otherwise LGTM

ice-tychon (approved):

PR #1780, opened 2025-08-01
fix: android e2e background messages

ice-endymion:
## Description
- rewrite background handler logic for better stability
~- add one time read db provider for push notifications, where we don't need reactivity~

## Task ID
ION-3487

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/user/providers/user_delegation_provider.r.dart
+++ b/lib/app/features/user/providers/user_delegation_provider.r.dart
@@ -126,13 +126,3 @@ class UserDelegationFromDb extends _$UserDelegationFromDb {
     return null;
   }
 }
-
-// One-time read provider for push notifications and other one-off operations
-@riverpod
-Future<UserDelegationEntity?> userDelegationFromDbOnce(
-  Ref ref,
-  String masterPubkey,
-) async {
-  final userDelegationDao = ref.read(userDelegationDaoProvider);
-  return userDelegationDao.get(masterPubkey);
-}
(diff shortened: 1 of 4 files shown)


lib/app/features/user/providers/user_delegation_provider.r.dart
```diff
@@ -126,3 +126,13 @@ class UserDelegationFromDb extends _$UserDelegationFromDb {
...
+
+// One-time read provider for push notifications and other one-off operations
+@riverpod
+Future<UserDelegationEntity?> userDelegationFromDbOnce(
+  Ref ref,
+  String masterPubkey,
+) async {
+  final userDelegationDao = ref.read(userDelegationDaoProvider);
+  return userDelegationDao.get(masterPubkey);
+}
```
ice-kreios, line 138: feels like a bit of over-abstraction — maybe we could just use `userDelegationDao.get() `directly?

lib/app/features/user/providers/user_metadata_provider.r.dart
```diff
@@ -94,3 +94,13 @@ class UserMetadataFromDb extends _$UserMetadataFromDb {
...
+
+// One-time read provider for push notifications and other one-off operations
+@riverpod
+Future<UserMetadataEntity?> userMetadataFromDbOnce(
+  Ref ref,
+  String masterPubkey,
+) async {
+  final userMetadataDao = ref.read(userMetadataDaoProvider);
+  return userMetadataDao.get(masterPubkey);
+}
```
ice-kreios, line 106: feels like a bit of over-abstraction — maybe we could just use userMetadataDao.get() directly?

ice-kreios (approved):

ice-tychon (approved):

ice-brontes (approved):

PR #1789, opened 2025-08-05
fix: poll vote restrictions

ice-erebus:
## Description
- Fix percentage counting on UI side
- Block vote request if it's already loading 

## Task ID
ION-3506

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

IMG_6625

### DIFF
--- a/lib/app/features/feed/polls/view/components/poll_vote_result_item.dart
+++ b/lib/app/features/feed/polls/view/components/poll_vote_result_item.dart
@@ -72,7 +72,8 @@ class PollResultItem extends HookWidget {
                 alignment: AlignmentDirectional.centerStart,
                 child: Container(
                   height: 34.0.s,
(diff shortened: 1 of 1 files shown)


lib/app/features/feed/polls/view/components/poll_vote_result_item.dart
```diff
@@ -27,8 +27,7 @@ class PollResultItem extends HookWidget {
       },
       [votes, totalVotes],
     );
-
-    final percentageInt = (percentage * 100).toInt();
+    final percentageInt = (percentage * 100).round();
```
ice-kreios, line 30: > votes: 1
> totalVotes: 3
> 

> 
> (1 / 3) * 100 = 33.333... → .round() = 33
> Sum of displayed percentages = 33 + 33 + 33 = 99

In this case, we won’t show the total percentage of 100. Do you agree with me?

PR #1831, opened 2025-08-11
fix: apply border radius for the network icons

ice-brontes:
## Description
1. Applied border radius for the network icons.
2. Created general widget/type for networks and coins.
3. Made a dynamic mechanism to calculate `borderRadius` according to the widget size. 

## Task ID
ION-3592

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactoring
- [ ] Documentation
- [ ] Chore

## Demo
https://github.com/user-attachments/assets/ee17a722-10a6-4739-a823-feaac246441c


lib/app/components/icons/coin_icon.dart
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:flutter/material.dart';
+import 'package:ion/app/components/icons/wallet_item_icon_type.dart';
+import 'package:ion/app/components/icons/wallet_item_icon_widget.dart';
+
+class CoinIconWidget extends StatelessWidget {
```
ice-kreios, line 7: asking to be sure - do we need this widget because we might use `NetworkIconWidget`. wdyt? otherwise LGTM

`CoinIconWidget = NetworkIconWidget(color:null)`

ice-kreios (approved):

ice-tychon (approved):

ice-lycurgus (approved):

PR #1844, opened 2025-08-12
fix: who can reply flow logic

ice-hades:
## Description
- Fixed an issue when `Who can reply` restricted warning was preliminary shown but rather data is not fully loaded yet;
- Fixed issue when share a post via a story of a verified user lacked the verified badge;
- Removed caching from can reply provider;

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Chore

### DIFF
--- a/lib/app/components/counter_items_footer/replies_counter_button.dart
+++ b/lib/app/components/counter_items_footer/replies_counter_button.dart
@@ -34,13 +34,13 @@ class RepliesCounterButton extends HookConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final repliesCount = ref.watch(repliesCountProvider(eventReference));
     final isReplied = ref.watch(isRepliedProvider(eventReference));
-    final canReply = ref.watch(canReplyProvider(eventReference)).valueOrNull;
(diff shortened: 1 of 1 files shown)


ice-lycurgus (dismissed):

ice-kreios (dismissed):

lib/app/components/counter_items_footer/replies_counter_button.dart
```diff
@@ -34,21 +34,22 @@ class RepliesCounterButton extends HookConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final repliesCount = ref.watch(repliesCountProvider(eventReference));
     final isReplied = ref.watch(isRepliedProvider(eventReference));
-    final canReply = ref.watch(canReplyProvider(eventReference)).valueOrNull ?? false;
+    final canReply = ref.watch(canReplyProvider(eventReference)).valueOrNull ?? true;
```
ice-lycurgus, line 37: Could we do instead of defaulting to true on null (that allows potentially restricted 'reply'), explicitly show loading state until real value is fetched?

I mean consider disabling reply action with spinner before we know true/false? Or switch back to 'false' you have before.

PR #1857, opened 2025-08-13
fix: follow button not updating after unfollow in stories

ice-ajax:
## Description
This PR fixes an issue where unfollow button in stories viewer didn't change to `follow` after unfollowing.

## Task ID
3415

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/68d0ad62-4dc9-45f6-a5f2-6375c93d9cf6

### DIFF
--- a/lib/app/features/optimistic_ui/features/follow/follow_provider.r.dart
+++ b/lib/app/features/optimistic_ui/features/follow/follow_provider.r.dart
@@ -15,6 +15,7 @@ part 'follow_provider.r.g.dart';
 
 @riverpod
 OptimisticService<UserFollow> followService(Ref ref) {
+  keepAliveWhenAuthenticated(ref);
   final manager = ref.watch(followManagerProvider);
   final initialFollows = ref.read(currentUserSyncFollowListProvider)?.masterPubkeys ?? [];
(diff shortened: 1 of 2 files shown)


lib/app/features/optimistic_ui/features/follow/follow_provider.r.dart
```diff
@@ -25,6 +26,12 @@ OptimisticService<UserFollow> followService(Ref ref) {
   return service;
 }
 
+@riverpod
```
ice-kreios, line 29: i think we do not need `followGet` provider, because it is basically call another provider without additional business logic. 

we can directly use it like this, wdyt?

`var current = ref.read(followServiceProvider.notifier).get(pubkey);`

ice-kreios (dismissed):

lib/app/features/optimistic_ui/features/follow/follow_provider.r.dart
ice-ajax: fixed

ice-kreios (approved):

ice-lycurgus (approved):

PR #1869, opened 2025-08-15
fix(feed): fix content creation blocker dialog behavior

ice-tychon:
### What
- Re-enabled the dialog that was blocking content creation
- Also removed the NavigationAppBar for this dialog, because closing it caused an error – it couldn’t perform a pop.
– In addition, I created a function that invalidates CurrentUserMetadata to fetch up-to-date data.

### Why
The backend is now working again, so this functionality is no longer a blocker.

## Type of Change
- [x] Bug fix

## Screenshots
image

### DIFF
--- a/lib/app/features/feed/views/pages/feed_main_modal/feed_main_modal_page.dart
+++ b/lib/app/features/feed/views/pages/feed_main_modal/feed_main_modal_page.dart
@@ -37,16 +37,15 @@ class FeedMainModalPage extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final hasNftCollectionState = ref.watch(hasIonContentNftCollectionProvider);
-    final body = hasNftCollectionState.whenOrNull(
-          data: (hasNftCollection) =>
-              hasNftCollection ? const _CreateContentModal() : const _ContentCreationBlockedModal(),
-        ) ??
-        const _CreateContentLoadingModal();
 
     return SheetContent(
       backgroundColor: context.theme.appColors.secondaryBackground,
       topPadding: 0.0.s,
-      body: body,
+      body: hasNftCollectionState.maybeWhen(
+        data: (hasNftCollection) =>
+            hasNftCollection ? const _CreateContentModal() : const _ContentCreationBlockedModal(),
+        orElse: () => const _CreateContentLoadingModal(),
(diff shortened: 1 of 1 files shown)


lib/app/features/feed/views/pages/feed_main_modal/feed_main_modal_page.dart
```diff
@@ -36,14 +36,12 @@ class FeedMainModalPage extends ConsumerWidget {
...
-    //           hasNftCollection ? const _CreateContentModal() : const _ContentCreationBlockedModal(),
-    //     ) ??
-    //     const _CreateContentLoadingModal();
-    const body = _CreateContentModal();
+    final hasNftCollectionState = ref.watch(hasIonContentNftCollectionProvider);
+    final body = hasNftCollectionState.whenOrNull(
+          data: (hasNftCollection) =>
+              hasNftCollection ? const _CreateContentModal() : const _ContentCreationBlockedModal(),
+        ) ??
+        const _CreateContentLoadingModal();
```
ice-kreios, line 44: A simple suggestion -  would be to use `maybeWhen` on `hasNftCollectionState` to determine the widget, instead of creating it inside the build method. This would align with the approach we’re already using on other screens. does it make sense?

```
return SheetContent(
    backgroundColor: context.theme.appColors.secondaryBackground,
    topPadding: 0.0.s,
    body: hasNftCollectionState.maybeWhen(
      data: (hasNftCollection) => hasNftCollection
          ? const _CreateContentModal()
          : const _ContentCreationBlockedModal(),
      orElse: () => const _CreateContentLoadingModal(),
    ),
  );
```

lib/app/features/user/providers/user_metadata_provider.r.dart
```diff
@@ -116,3 +116,24 @@ class UserMetadataFromDb extends _$UserMetadataFromDb {
     return null;
   }
 }
+
+Future<void> invalidateCurrentUserMetadataProviders(
+  WidgetRef ref, {
+  ActionType? actionType,
+}) async {
+  final pubkey = ref.read(currentPubkeySelectorProvider);
```
ice-kreios, line 124: i think we should use `watch` instead of `read`. wdyt?

ice-brontes (approved):

PR #1906, opened 2025-08-20
fix: relays not found after foreground

ice-hades:
## Description
- Fixed an issue when user get `Relay not found` after foreground;
Main idea is to pause scheduled relays ranking when in the background and continue when is in foreground;

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/services/pauseable_periodic_runner/pauseable_periodic_runner.r.dart
+++ b/lib/app/services/pauseable_periodic_runner/pauseable_periodic_runner.r.dart
@@ -4,6 +4,7 @@ import 'dart:async';
 import 'dart:ui';
 
 import 'package:dio/dio.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
 import 'package:ion/app/features/core/providers/app_lifecycle_provider.r.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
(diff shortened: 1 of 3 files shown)


lib/app/services/pauseable_periodic_runner/pauseable_periodic_runner.r.dart
```diff
@@ -0,0 +1,138 @@
...
+
+part 'pauseable_periodic_runner.r.g.dart';
+
+/// A reusable periodic runner that:
+/// - Fires a callback on a fixed [interval]
+/// - Pauses when app goes to background/inactive and remembers remaining time
+/// - On resume, schedules a one-shot for max(remaining, [minResumeDelay]), then returns to periodic cadence
+/// - Cancels and rotates a [CancelToken] for in-flight async work on each pause/resume and tick
+@Riverpod(keepAlive: true)
+class PauseablePeriodicRunner extends _$PauseablePeriodicRunner {
```
ice-tychon, line 18: I'd avoid it to be a Riverpod Notifier provider. It's better to create a good old Dart class, and use Riverpod just to create (and dispose) it's instance.

ice-kreios (approved):

PR #1928, opened 2025-08-22
fix: filter counter results by event id

ice-orion:
## Description
This PR fixes

1. Incorrect counters REQ filter for 5400 results - 6400/7000.
Previously, we requested all 6400 and 7000 events for the current user pubkey while waiting for the result of 5400. This caused the backend to return all stored 6400 and 7000 events for the given user.
This PR changes the behavior to request only the 6400 and 7000 events related to the specific 5400 event.

2. Redundant requesting of stories counters.
Previously, we requested total number of stories in a loop for each story. The PR changes it to request it only once for each uniq story author.

## Task ID
ION-3719

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/views/pages/feed_page/components/stories/components/story_list_item.dart
+++ b/lib/app/features/feed/views/pages/feed_page/components/stories/components/story_list_item.dart
@@ -27,7 +27,7 @@ class StoryListItem extends HookConsumerWidget {
 
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    // prefetch stories counter for the giver author
+    // prefetch stories counter for the given author
     ref.watch(storiesCountProvider(pubkey));
     final userMetadata = ref.watch(cachedUserMetadataProvider(pubkey));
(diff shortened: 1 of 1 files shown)


lib/app/features/user/providers/count_provider.r.dart
```diff
@@ -79,11 +76,9 @@ class Count extends _$Count {
       ..addFilter(
         RequestFilter(
           kinds: const [EventCountResultEntity.kind, EventCountErrorEntity.kind],
-          tags: addCurrentPubkey
-              ? {
-                  '#p': [currentPubkey],
-                }
-              : null,
+          tags: {
+            '#e': [requestEvent.id],
```
ice-kreios, line 80: if we do not use `currentPubkey` anymore we might delete the related block wdyt?

```
final currentPubkey = ref.read(currentPubkeySelectorProvider);
    if (currentPubkey == null) {
      throw UserMasterPubkeyNotFoundException();
    }
```

ice-kreios (dismissed):

PR #1970, opened 2025-09-01
feat: navigate to post details on reply tap on media full screen page

ice-ajax:
## Description
This PR changes the behaviour of reply button in full screen media pages so that it navigates to post/article detail pages.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

https://github.com/user-attachments/assets/d5d854ec-b5ec-4b37-886a-c46fd04e7a24

### DIFF
--- a/lib/app/components/counter_items_footer/replies_counter_button.dart
+++ b/lib/app/components/counter_items_footer/replies_counter_button.dart
@@ -23,12 +23,14 @@ class RepliesCounterButton extends HookConsumerWidget {
     required this.eventReference,
     this.color,
     this.padding,
+    this.onTap,
     super.key,
   });
 
   final EventReference eventReference;
   final Color? color;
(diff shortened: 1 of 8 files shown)


lib/app/components/counter_items_footer/replies_counter_button.dart
```diff
@@ -40,6 +42,10 @@ class RepliesCounterButton extends HookConsumerWidget {
     return GestureDetector(
       behavior: HitTestBehavior.opaque,
       onTap: () async {
+        if (onTap != null) {
```
ice-evander, line 45: You can use 'if case' to avoid bang operator:
```
if (onTap case final onTap?) {
   onTap();
   return;
}
```

ice-evander (approved):

PR #1987, opened 2025-09-02
fix(wallet): nfts reloaded each time user opens nfts tag

ice-tychon:
## Description
Previously, the “NFTs” tab was reloaded every time the user opened it. This happened because the `currentNftsProvider` and `walletNftsViewModelProvider` were disposed of when the widget was disposed.

## Type of Change
- [x] Bug fix

## Screenshots 
https://github.com/user-attachments/assets/ad668913-b974-4572-8a66-864608bbad4a

### DIFF
--- a/lib/app/features/wallets/providers/current_nfts_provider.r.dart
+++ b/lib/app/features/wallets/providers/current_nfts_provider.r.dart
@@ -8,7 +8,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'current_nfts_provider.r.g.dart';
 
-@riverpod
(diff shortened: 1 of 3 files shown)


lib/app/features/wallets/providers/current_nfts_provider.r.dart
```diff
@@ -8,7 +8,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'current_nfts_provider.r.g.dart';
 
-@riverpod
+@Riverpod(keepAlive: true)
```
ice-kreios, line 11: should we dispose it when user logged out?

PR #1991, opened 2025-09-03
fix: replace ffmpeg with pure dart implementation for webp to jpeg conversion on background push notification handler isolate

ice-ajax:
## Description
This PR fixes an issue where android push notifications didn't contain any images an icons when push was received when app was in background

## Additional Notes
The main issue here is that we use ffmpegKit for converting push notification images from webp to jpeg. However this library [doesn't work on other isolates than main one](https://github.com/arthenica/ffmpeg-kit/issues/367). However firebase background notification handler runs on a background isolate. 
The solution uses a pure dart solution for image conversion that works on all isolates.

## Task ID
3570

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

ScreenShot 2025-09-03 at 13 02 12@2x

### DIFF
--- a/lib/app/services/local_notifications/local_notifications.r.dart
+++ b/lib/app/services/local_notifications/local_notifications.r.dart
@@ -7,10 +7,9 @@ import 'dart:io';
 import 'package:dio/dio.dart';
 import 'package:flutter_local_notifications/flutter_local_notifications.dart';
 import 'package:hooks_riverpod/hooks_riverpod.dart';
-import 'package:ion/app/services/compressors/compress_executor.r.dart';
 import 'package:ion/app/services/compressors/image_compressor.r.dart';
+import 'package:ion/app/services/converters/dart_webp_to_jpeg_converter.dart';
 import 'package:ion/app/services/logger/logger.dart';
-import 'package:ion/app/services/media_service/media_service.m.dart';
 import 'package:ion/app/services/uuid/uuid.dart';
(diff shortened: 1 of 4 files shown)


lib/app/services/local_notifications/local_notifications.r.dart
```diff
@@ -217,14 +216,10 @@ class LocalNotificationsService {
       final tempFile = File(tempFilePath);
       await tempFile.writeAsBytes(data);
 
-      // Need to compress from webp to jpg since conversations push notifications not support webp
-      final compressor = ImageCompressor(compressExecutor: CompressExecutor());
-      final compressedMedia = await compressor.compress(
-        MediaFile(path: tempFilePath),
-        to: ImageCompressionType.jpeg,
```
ice-kreios, line 224: if we do not need anymore `ImageCompressionType.jpeg` to clean up the code we might delete the below enum wdyt?

```

enum ImageCompressionType {
  webp,
  jpeg;

  String get mimeType => switch (this) {
        webp => MimeType.image.value,
        // We use this for push notifications only
        jpeg => LocalMimeType.jpeg.value,
      };
}
```

ice-ajax: it is still used inside the existing ImageCompressor

PR #2013, opened 2025-09-08
feat: add abbreviations for big crypto balances

ice-brontes:
## Description
Starting from millions, add the abbreviation for the big crypto values:
1,200,000	              -  1.2M
18,308,397	              -  18.3M
508,920,000	              -  508.9M
7,320,000,000	      -  7.32B
18,308,397,101,537.31  -  18.308T

## Task ID
ION-3692

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/wallets/views/utils/crypto_formatter.dart
+++ b/lib/app/features/wallets/views/utils/crypto_formatter.dart
@@ -36,16 +36,6 @@ String formatCrypto(double value, [String? currency]) {
   return formatted;
 }
 
-String _processDecimalPart(String decimalPart) {
-  // Truncate to maximum 3 decimal places
-  final truncated = decimalPart.length > 3 ? decimalPart.substring(0, 3) : decimalPart;
-
-  // Remove trailing zeros
-  final trimmed = truncated.replaceAll(RegExp(r'0+$'), '');
-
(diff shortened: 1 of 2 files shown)


lib/app/features/wallets/views/utils/crypto_formatter.dart
```diff
@@ -2,9 +2,23 @@
...
+const _million = 1000000.0;
+const _billion = 1000000000.0;
+const _trillion = 1000000000000.0;
+
+// Scale information for abbreviation formatting
+const List<({double value, String suffix})> _scaleInfo = [
+  (value: _trillion, suffix: 'T'),
+  (value: _billion, suffix: 'B'),
+  (value: _million, suffix: 'M'),
+];
```
ice-kreios, line 16: i think we should do locale-aware formatting for these suffixes since 'T', 'B', 'M' may not be appropriate for all regions/languages. Also previously as i mentioned we are still using en_US locale for decimal formatting for all users, does it make sense to check these cases before production release?

ice-kreios (approved):

PR #2032, opened 2025-09-11
fix: initial likes from cache number

ice-cerberus:
## Description
 - Only fall back to counting cached reactions if no counter entity exists

## Task ID
ION-3826

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/optimistic_ui/features/likes/post_like_provider.r.dart
+++ b/lib/app/features/optimistic_ui/features/likes/post_like_provider.r.dart
@@ -28,8 +28,7 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'post_like_provider.r.g.dart';
 
-@riverpod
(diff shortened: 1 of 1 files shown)


ice-tychon (dismissed):

ios/Podfile.lock
```diff
@@ -102,6 +102,8 @@ PODS:
     - BNBSdkCore (= 1.17.5)
   - camera_avfoundation (0.0.1):
     - Flutter
+  - cryptography_flutter (0.2.0):
+    - Flutter
```
ice-kreios, line 106: i think we should update the Podfile.lock because it has been deleted with https://github.com/ice-blockchain/ion-framework/pull/2033

ice-brontes (approved):

PR #2050, opened 2025-09-14
fix: after sending snow to another user transaction disappears from coin details

ice-brontes:
## Description
1. Add unknown token type of transactions to the database.
2.  Add `UndefinedTransactionsBinder` service to bind unknown token transactions with coin id by contract address from wallet view.
3. Handle different tokens with the same symbol in the wallet view aggregation.

## Task ID
ION-3710

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/wallets/data/database/wallets_database.m.dart
```diff
@@ -237,6 +237,18 @@ class WalletsDatabase extends _$WalletsDatabase {
...
+          final columnExists = await isColumnExists(
+            tableName: schema.transactionsTableV2.actualTableName,
+            columnName: 'asset_contract_address',
+          );
+          if (!columnExists) {
+            await m.addColumn(
+              schema.transactionsTableV2,
+              schema.transactionsTableV2.assetContractAddress,
+            );
+          }
```
ice-kreios, line 250: i think we do not need to check `isColumnExist` because we add `asset_contract_address` col with v20 and v19 does not contain it. wdyt?

ice-kreios (approved):

PR #2084, opened 2025-09-17
fix: video compression speed

ice-cerberus:
## Description
- replaced libx264 with h264

## Task ID
ION-3883

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/packages/ion_identity_client/lib/src/signer/password_signer.dart
+++ b/packages/ion_identity_client/lib/src/signer/password_signer.dart
@@ -224,7 +224,7 @@ class PasswordSigner {
     final localAuth = LocalAuthentication();
 
     final results = await Future.wait<bool>([
-      Future.value(true),
+      localAuth.canCheckBiometrics,
       localAuth.isDeviceSupported(),
     ]);
 


lib/app/services/media_service/ffmpeg_commands_config.dart
```diff
@@ -24,7 +24,7 @@ class FFmpegCommands {
       videoCodec,
       '-preset',
       preset,
-      '-maxrate',
+      '-b:v',
```
ice-kreios, line 27: I’m not sure forcing the actual video bitrate is a good idea. If the original video’s bitrate is lower than the target, the file size will just increase without any quality gain. I think it makes more sense to raise the max bitrate limit instead. What do you think?

ice-kreios (dismissed):

lib/app/services/compressors/video_compressor.r.dart
```diff
@@ -96,16 +81,30 @@ class VideoCompressor implements Compressor<VideoCompressionSettings> {
...
+      final originalBitrate = await getVideoBitrate(file.path);
+      final targetBitrateValue = settings.maxRate.bitrate;
+      final shouldCompress = originalBitrate == null || originalBitrate > targetBitrateValue * 1000;
+      if (!shouldCompress) {
+        final dimensions = await getVideoDimensions(file.path);
+        return file.copyWith(
+          width: dimensions.width,
+          height: dimensions.height,
+        );
+      }
```
ice-kreios, line 100: I don’t think this is valid because we should always compress and generate an MP4 file. With these changes, if a user uploads a video in a different format with a bitrate lower than 4000, it won’t be processed.

lib/app/services/compressors/video_compressor.r.dart
```diff
@@ -40,32 +39,18 @@ class VideoCompressionSettings {
...
-
   static const highQuality = VideoCompressionSettings(
-    videoCodec: FFmpegVideoCodecArg.libx264,
-    preset: FfmpegPresetArg.slow,
-    maxRate: FfmpegBitrateArg.high,
+    videoCodec: FFmpegVideoCodecArg.h264,
+    maxRate: FfmpegBitrateArg.medium,
     bufSize: FfmpegBitrateArg.highest,
-    scale: FfmpegScaleArg.p1080,
+    scale: FfmpegScaleArg.truncateOnly,
```
ice-kreios, line 46: As a technical requirement, we decided to always export videos in 1080p.
	1.	We should inform the PO and make the decision together.
	2.	If a user uploads an 8K video, it won’t be resized as I see.

We should test this with videos of different resolutions

PR #2143, opened 2025-09-25
chore: add networks comparator

ice-brontes:
## Description
Added the networks comparator to display networks in the correct order in the receive nft flow.

## Task ID
ION-3993

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

Simulator Screenshot - iPhone 16 Pro - 2025-09-25 at 12 58 29


ice-tychon (dismissed):

lib/app/features/wallets/domain/networks/networks_comparator.r.dart
```diff
@@ -0,0 +1,84 @@
...
+    final index = _nftPopularityList.indexWhere(
+      (ids) => ids.mainnetNetworkId == networkId || ids.testnetNetworkId == networkId,
+    );
+    return index == -1 ? -1 : index;
+  }
+
+  @override
+  int compare(NetworkData networkA, NetworkData networkB) {
+    // 0. ION always comes first, regardless of other conditions
+    final isNetworkATop = _topNetwork.matches(networkA);
```
ice-evander, line 40: Does it makes sense to handle _topNetwork here separately, if we have it as first item in the _nftPopularityList? I believe it we should chose only one of the approaches: 
either to have this _topNetwork logic here and remove it from the list
OR
to remove this isNetworkXTop logic and leave it to be handle by the general approach started at line 47
Wdyt?

ice-kreios (approved):

ice-tychon (approved):

PR #2168, opened 2025-09-27
chore: content labeler flow resilience

ice-hades:
## Description
- Some changes aimed to improve content labeler flow resilience;

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [x] Chore

### DIFF
--- a/lib/app/components/text_editor/utils/debounced_quill_controller_listener.dart
+++ b/lib/app/components/text_editor/utils/debounced_quill_controller_listener.dart
@@ -8,26 +8,22 @@ Stream<String> debouncedQuillControllerListener(
   QuillController controller, {
   Duration duration = const Duration(milliseconds: 500),
 }) {
-  late StreamController<String> streamController;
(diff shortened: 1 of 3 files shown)


ice-evander (approved):

lib/app/components/text_editor/utils/debounced_quill_controller_listener.dart
```diff
@@ -11,16 +11,26 @@ Stream<String> debouncedQuillControllerListener(
   late StreamController<String> streamController;
   Timer? debounce;
 
+  late void Function() listener;
```
ice-tychon, line 14: Can we avoid the `late` keyword? Current code is simple now, but can lead to some issues in future.

ice-tychon: `StreamController<String> streamController` is also `late`

ice-evander (approved):

PR #2220, opened 2025-10-03
fix: make sure there is only one request filter

ice-hades:
## Description
With more than 1 request filter in pagination requests there can be a situation
when we get created at of the last event in the stream but it is correct only for one of the filters.
So here I make sure we always has only one filter for pagination requests to fix such scenarios.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/feed/providers/replies_data_source_provider.r.dart
+++ b/lib/app/features/feed/providers/replies_data_source_provider.r.dart
@@ -39,35 +39,23 @@ List<EntitiesDataSource>? repliesDataSource(
           (entity is ModifiablePostEntity &&
               entity.data.parentEvent?.eventReference == eventReference) ||
           (entity is PostEntity && entity.data.parentEvent?.eventReference == eventReference),
-      requestFilters: [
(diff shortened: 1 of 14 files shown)


lib/app/features/feed/providers/replies_data_source_provider.r.dart
```diff
@@ -39,35 +39,23 @@ List<EntitiesDataSource>? repliesDataSource(
...
-              ExpirationSearchExtension(expiration: false),
-            ],
-          ).toString(),
-          limit: 10,
-        ),
-      ],
+      requestFilter: RequestFilter(
+        kinds: const [ModifiablePostEntity.kind, PostEntity.kind],
+        tags: Map.fromEntries([relatedTags]),
+        search: SearchExtensions(
```
ice-priapus, line 45: is it fine that previously filters have different search extension and now it's same list of extensions for both kinds?

PR #2222, opened 2025-10-03
fix: chat scroll jumps

ice-priapus:
## Description
Improvements for search in chat and for scroll in chat list 

## Task ID
ION-3966,ION-3984

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/search/providers/chat_search/chat_messages_search_provider.r.dart
+++ b/lib/app/features/search/providers/chat_search/chat_messages_search_provider.r.dart
@@ -29,16 +29,16 @@ Future<List<ChatSearchResultItem>?> chatMessagesSearch(Ref ref, String query) as
 
   final result = <ChatSearchResultItem>[];
 
-  entities.sortedBy((entity) => entity.createdAt.toDateTime).reversed.forEach((message) async {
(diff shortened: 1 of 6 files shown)


ice-damocles (approved):

lib/app/features/search/providers/chat_search/chat_messages_search_provider.r.dart
```diff
@@ -29,24 +29,24 @@ Future<List<ChatSearchResultItem>?> chatMessagesSearch(Ref ref, String query) as
 
   final result = <ChatSearchResultItem>[];
 
-  entities.sortedBy((entity) => entity.createdAt.toDateTime).reversed.forEach((message) async {
+  for (final message in entities.sortedBy((entity) => entity.createdAt.toDateTime).reversed) {
```
ice-kreios, line 32: i think it should be `publishedAt` instead of `createdAt`. because we sort messages by `publishedAt.` - you can test it with edited messages. do you agree with me @ice-damocles ?

PR #2256, opened 2025-10-07
chore: wallet fixes for converting the crypto amount

ice-brontes:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore


lib/app/features/wallets/utils/crypto_amount_converter.dart
```diff
@@ -0,0 +1,27 @@
...
+import 'package:decimal/decimal.dart';
+import 'package:ion/app/services/logger/logger.dart';
+
+/// Converts a blockchain-precise numeric string (e.g., `"1000000000000000000"`)
+/// into a human-readable Decimal value (e.g., `Decimal.parse('1.0')`).
+double fromBlockchainUnits(String input, int decimals) {
+  try {
+    final value = BigInt.parse(input);
+    final divisor = BigInt.from(10).pow(decimals);
+    return (Decimal.fromBigInt(value) / Decimal.fromBigInt(divisor)).toDouble();
```
ynlvko, line 12: Do we need to convert it to `BigInt` before using the `Decimal`? Why not use `Decimal.parse` rightaway?

ice-tychon, line 12: Do we need to convert it to `BigInt` before using the `Decimal`? Why not use `Decimal.parse` rightaway?

ice-tychon (approved):

lib/app/features/wallets/utils/crypto_amount_converter.dart
```diff
@@ -0,0 +1,27 @@
...
+  } on FormatException catch (_) {
+    Logger.error('Failed to parse coins amount with `$input` value.');
+    return 0;
+  }
+}
+
+/// Converts a human-readable double value (e.g., `1.23`)
+/// into a blockchain-precise BigInt (e.g., `BigInt.parse('1230000000000000000')`).
+BigInt toBlockchainUnits(double? amountValue, int decimals) {
```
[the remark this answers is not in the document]
ice-brontes: I decided to use `BigInt` here because this method should return values like `1230000000000000000`, not `1230000000000000000.999999999`, cos it's related only to crypto, so the format should be strict.

PR #2265, opened 2025-10-07
fix: followers loading more on scrolling

ice-hades:
## Description
- Fixed issue with follower pop up. Load more actually issued the same request and fetched the same batch of followers as the original request. Need to reuse the same List<EntitiesDataSource> for entitiesPagedDataProvider or every time this list is new a new provider instance is created

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

## Screenshots

https://github.com/user-attachments/assets/b176caa0-c60d-4713-b3a5-a54176f52232

### DIFF
--- a/lib/app/features/user/providers/followers_provider.r.dart
+++ b/lib/app/features/user/providers/followers_provider.r.dart
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: ice License 1.0
 
+import 'dart:async';
+
 import 'package:ion/app/features/core/model/paged.f.dart';
 import 'package:ion/app/features/ion_connect/providers/entities_paged_data_provider.m.dart';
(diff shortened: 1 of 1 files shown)


ice-evander (approved):

lib/app/features/user/providers/followers_provider.r.dart
```diff
@@ -11,6 +13,9 @@ part 'followers_provider.r.g.dart';
 
 @riverpod
 class Followers extends _$Followers {
```
ice-tychon, line 15: This class feels a bit cryptic to me. Do you think it could be refactored to remove the Riverpod Notifier provider without much hassle?

PR #3717, opened 2026-03-18
fix: fallback during recovery only for platform exception

ice-hades:
## Description
- Fallback to password flow during recovery only on platform exception;
- No error pop up in case of platform exception or cancel exception during recovery flow;
- - Fixed fallback on platform exception during register flow;

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

### DIFF
--- a/lib/app/features/auth/flows/run_sign_up_then_login.dart
+++ b/lib/app/features/auth/flows/run_sign_up_then_login.dart
@@ -22,7 +22,7 @@ Future<void> runSignUpThenLogin({
 }) async {
   String? capturedPassword;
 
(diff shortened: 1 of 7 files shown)


ice-tychon (approved):

ice-phobos (approved):

ice-priapus (commented):
As I see from the PR now if user cancels the passkey we will fallback. Is there a way for a user to go back to passkey was if he canceled it by accident?

ice-priapus (approved):

PR #3725, opened 2026-03-19
fix: enhance RelayChain and SwapQuoteInfo models with new httpRpcUrl

ice-erebus:
## Description
- Wait for approve txs confirmation from blockchain by itself

## Additional Notes
- NA

## Task ID
- ION-5844

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactoring
- [ ] Documentation
- [ ] Chore

- NA

### DIFF
--- a/packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
+++ b/packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:http/http.dart' as http;
+import 'package:ion_swap_client/exceptions/ion_swap_exception.dart';
+import 'package:web3dart/web3dart.dart';
+
+mixin WaitForConfirmationMixin {
+  Future<void> waitForConfirmation({
+    required String txHash,
+    required String rpcUrl,
+    int maxTries = 20,
+    Duration pollInterval = const Duration(seconds: 3),
+  }) async {
+    if (!txHash.startsWith('0x') || txHash.length != 66) {
+      return;
+    }
+
+    final httpClient = http.Client();
+    final web3Client = Web3Client(rpcUrl, httpClient);
+
+    try {
+      for (var i = 0; i < maxTries; i++) {
+        final receipt = await web3Client.getTransactionReceipt(txHash);
+
+        if (receipt != null) {
+          if (receipt.status ?? false) {
+            return;
+          }
+
+          throw IonSwapException(
+            'Transaction failed on-chain, tx hash: $txHash, status: ${receipt.status}, from: ${receipt.from}, to: ${receipt.to}, gasUsed: ${receipt.gasUsed}, effectiveGasPrice: ${receipt.effectiveGasPrice}',
+          );
+        }
+
+        await Future<void>.delayed(pollInterval);
+      }
+
+      throw IonSwapException(
+        'Timed out waiting for transaction confirmation, tx hash: $txHash, rpcUrl: $rpcUrl',
+      );
+    } finally {
+      await web3Client.dispose();
+      httpClient.close();
+    }
+  }
+}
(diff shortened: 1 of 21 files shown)


packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:http/http.dart' as http;
+import 'package:ion_swap_client/exceptions/ion_swap_exception.dart';
+import 'package:web3dart/web3dart.dart';
+
+mixin WaitForConfirmationMixin {
+  Future<void> waitForConfirmation({
```
ice-tychon, line 8: Can we reuse `withRetry`?

ice-tychon (approved):

packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
ice-erebus: Yes, we can but in such case I need to move it in separate package or copy it in mine as it's in `withRetry` located in `app package` and `waitForConfirmation` inside `ion_swap_client`

I think it's too much for one place

packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
```diff
@@ -0,0 +1,46 @@
...
+    try {
+      for (var i = 0; i < maxTries; i++) {
+        final receipt = await web3Client.getTransactionReceipt(txHash);
+
+        if (receipt != null) {
+          if (receipt.status ?? false) {
+            return;
+          }
+
+          throw IonSwapException(
```
ice-priapus, line 30: is it worth make it more general, not only for swap? If so we could move it from swap client package and use IonIdentityTransactionException instead of IonSwapException

packages/ion_swap_client/lib/services/dex_service.dart
```diff
@@ -96,17 +104,25 @@ class DexService {
...
-            address,
-            approveTx.dexContractAddress,
-            wallet.id,
-          );
-
-          if (allowance2 < bigIntAmount) {
-            throw IonSwapException('Failed to approve token allowance, tx hash: $txHash');
+          final rcpUrl = swapQuoteInfo.sellBlockchainRcpUrl;
+          if (rcpUrl == null) {
+            await Future<void>.delayed(SwapConstants.delayAfterApproveDuration);
```
ice-priapus, line 109: so if rcpURL is null we are not having allowance check. is it ok?

packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
```diff
@@ -0,0 +1,46 @@
...
+import 'package:web3dart/web3dart.dart';
+
+mixin WaitForConfirmationMixin {
+  Future<void> waitForConfirmation({
+    required String txHash,
+    required String rpcUrl,
+    int maxTries = 20,
+    Duration pollInterval = const Duration(seconds: 3),
+  }) async {
+    if (!txHash.startsWith('0x') || txHash.length != 66) {
```
ice-priapus, line 14: don't we want to raise an exception in this case?

packages/ion_swap_client/lib/mixins/wait_for_confirmation_mixin.dart
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: ice License 1.0
+
+import 'package:http/http.dart' as http;
+import 'package:ion_swap_client/exceptions/ion_swap_exception.dart';
+import 'package:web3dart/web3dart.dart';
+
+mixin WaitForConfirmationMixin {
+  Future<void> waitForConfirmation({
+    required String txHash,
+    required String rpcUrl,
```
ice-priapus, line 10: is some places i can see rpc and others rcp. is this a typo or two different things?

